PR #3, opened 2015-08-19
Make the private key repository scanning actually work

jbunton:


atlassian_jwt_auth/key.py
```diff
@@ -120,8 +116,8 @@ def __init__(self, path):
         self.path = path
 
     def find_valid_key_ids(self, issuer):
-        issuer_directory = os.path.join(self.private_key_repository, issuer)
-        for filename in sorted(os.path.listdir(issuer_directory)):
+        issuer_directory = os.path.join(self.path, issuer)
+        for filename in sorted(os.listdir(issuer_directory)):
```
dbaxa: Can we reverse the sorting order ?

atlassian_jwt_auth/tests/test_signer_private_key_repo.py
```diff
@@ -0,0 +1,40 @@
...
+        self.assertEqual(expected_key_id, key_identifier.key_id)
+
+    def test_fails_if_issuer_has_no_valid_keys(self):
+        signer = self.create_signer_for_issuer('invalid-issuer')
+        with self.assertRaisesRegexp(Exception, 'Issuer has no valid keys'):
+            signer.generate_jwt('audience')
+
+    def test_fails_if_issuer_does_not_exist(self):
+        signer = self.create_signer_for_issuer('this-does-not-exist')
+        with self.assertRaisesRegexp(OSError, 'No such file or directory'):
```
dbaxa: We don't really need to use `assertRaisesRegexp`, `assertRaises` is fine.

jbunton: There are other kinds of OSError, in these cases the test should fail.

atlassian_jwt_auth/key.py
jbunton: Done.

atlassian_jwt_auth/key.py
```diff
@@ -97,21 +97,22 @@ def load(self, issuer):
...
+        for key_identifier in self.repository.find_valid_key_ids(issuer):
+            pass
+
+        if key_identifier:
             return key_identifier
-        raise Exception(
-            'Issuer has no valid keys in directory: ' %
-            issuer)
+        else:
+            raise Exception('Issuer has no valid keys: %s' % issuer)
```
dbaxa: Can you change this to a IOError or something else? - throwing Exception is poor practice.

atlassian_jwt_auth/key.py
```diff
@@ -97,21 +97,22 @@ def load(self, issuer):
 class FilePrivateKeyRetriever(object):
 
     def __init__(self, private_key_repository_path):
-        self.private_key_repository = FilePrivateKeyRepository(
```
dbaxa: Can we avoid renaming private_key_repository to repository ?

atlassian_jwt_auth/key.py
```diff
@@ -97,21 +97,22 @@ def load(self, issuer):
...
         key_identifier = self._find_first_key_id(issuer)
-        private_key_pem = self.private_key_repository.load_key(key_identifier)
+        private_key_pem = self.repository.load_key(key_identifier)
         return key_identifier, private_key_pem
 
     def _find_first_key_id(self, issuer):
-        for key_identifier in self.private_key_repository.find_valid_key_id(
-                issuer):
+        key_identifier = None
+        for key_identifier in self.repository.find_valid_key_ids(issuer):
```
dbaxa: instead of `pass` in the loop you can just return `return key_identifier`

atlassian_jwt_auth/key.py
jbunton: I suppose. The shorter name meant less split lines later on though.

atlassian_jwt_auth/key.py
jbunton: I renamed the method to `_find_last_key_id`, hopefully it's clearer now what it's doing.

atlassian_jwt_auth/tests/test_signer_private_key_repo.py
```diff
@@ -0,0 +1,40 @@
...
+        token = signer.generate_jwt('audience')
+        self.assertIsNotNone(token)
+
+    def test_picks_last_valid_key_id(self):
+        signer = self.create_signer_for_issuer('issuer-with-many-keys')
+        token = signer.generate_jwt('audience')
+        key_identifier = key._get_key_id_from_jwt_header(token)
+
+        expected_key_id = 'issuer-with-many-keys/key3.pem'
+        self.assertEqual(expected_key_id, key_identifier.key_id)
```
dbaxa: iirc the other test code does `self.assertEqual(actual, expected)`

atlassian_jwt_auth/key.py
jbunton: Ack, forgot about this. I don't think IOError is the right choice but I'll think of something.

atlassian_jwt_auth/tests/test_signer_private_key_repo.py
jbunton: I always get this wrong. There is some language/tool the other way around. Maybe I'm thinking of Jasmine?

Fixed.

atlassian_jwt_auth/tests/test_signer_private_key_repo.py
dbaxa: I guess.

atlassian_jwt_auth/key.py
dbaxa: I don't mind the line splitting.

atlassian_jwt_auth/key.py
```diff
@@ -97,21 +97,22 @@ def load(self, issuer):
...
+        private_key_pem = self.repository.load_key(key_identifier)
         return key_identifier, private_key_pem
 
-    def _find_first_key_id(self, issuer):
-        for key_identifier in self.private_key_repository.find_valid_key_id(
-                issuer):
+    def _find_last_key_id(self, issuer):
+        key_identifier = None
+        for key_identifier in self.repository.find_valid_key_ids(issuer):
+            pass
```
dbaxa: You can change the pass to `return key_identifier`

atlassian_jwt_auth/key.py
dbaxa: IOError is available in python 2 and 3 and IMHO it isn't a bad choice.

atlassian_jwt_auth/key.py
jbunton: I've changed it back.

atlassian_jwt_auth/key.py
jbunton: Done.

PR #20, opened 2016-03-03
Add support for obtaining a key identifier and private key from a dat uri.

dbaxa:


dbaxa:
@jbunton because there is no need to hit the exception.

Also, for now I am not going to create a new method in signer for the `DataUriPrivateKeyRetriever`.

PR #22, opened 2016-05-09
Make the generate claims kwargs explicit

SamWhited:
Currently you can't actually look at the function definition for `_generate_claims` or `generate_jwt` and see what the API is. This makes it (and one or two other kwargs) explicit.

It is still backwards and forwards compatible.

### DIFF
--- a/atlassian_jwt_auth/signer.py
+++ b/atlassian_jwt_auth/signer.py
@@ -9,11 +9,15 @@ from atlassian_jwt_auth import key
 
 class JWTAuthSigner(object):
 
-    def __init__(self, issuer, private_key_retriever, **kwargs):
+    def __init__(self, issuer, private_key_retriever,
+                 lifetime=None, algorithm='RS256'):
         self.issuer = issuer
         self.private_key_retriever = private_key_retriever
-        self.lifetime = kwargs.get('lifetime', datetime.timedelta(hours=1))
-        self.algorithm = kwargs.get('algorithm', 'RS256')
+        if lifetime is None:
+            self.lifetime = datetime.timedelta(hours=1)
+        else:
+            self.lifetime = lifetime
+        self.algorithm = algorithm
 
         if self.algorithm not in set(
                 algorithms.get_permitted_algorithm_names()):
@@ -23,7 +27,7 @@ class JWTAuthSigner(object):
             raise ValueError("lifetime, '%s',exceeds the allowed 1 hour max" %
                              (self.lifetime))
 
-    def _generate_claims(self, audience, **kwargs):
+    def _generate_claims(self, audience, additional_claims=None):
         """ returns a new dictionary of claims. """
         now = self._now()
         claims = {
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/signer.py
```diff
@@ -9,11 +9,15 @@
 
 class JWTAuthSigner(object):
 
-    def __init__(self, issuer, private_key_retriever, **kwargs):
+    def __init__(self, issuer, private_key_retriever,
```
dbaxa: I prefer the current approach of using `**kwargs`. What I feel is lacking is docstring to explain what `kwargs` can be used.

atlassian_jwt_auth/signer.py
```diff
@@ -23,7 +27,7 @@ def __init__(self, issuer, private_key_retriever, **kwargs):
             raise ValueError("lifetime, '%s',exceeds the allowed 1 hour max" %
                              (self.lifetime))
 
-    def _generate_claims(self, audience, **kwargs):
+    def _generate_claims(self, audience, additional_claims=None):
```
dbaxa: This is an internal method and I'd prefer to keep `kwargs`

atlassian_jwt_auth/signer.py
```diff
@@ -36,18 +40,22 @@ def _generate_claims(self, audience, **kwargs):
...
-        claims.update(kwargs.get('additional_claims', {}))
+        if additional_claims is not None:
+            claims.update(additional_claims)
         return claims
 
     def _now(self):
         return datetime.datetime.utcnow()
 
-    def generate_jwt(self, audience, **kwargs):
+    def generate_jwt(self, audience, additional_claims=None):
```
dbaxa: Switching this to have `additional_claims` instead of `kwargs` here certainly does seem like an improvement here.

atlassian_jwt_auth/signer.py
SamWhited: I don't understand the benefit of using kwargs unpacking? As far as I can tell it just makes things less explicit and makes it harder to see at a glance what's going on.

atlassian_jwt_auth/signer.py
SamWhited: Same here; what's the benefit to using kwargs unpacking?

dbaxa: Future internal api changes.

atlassian_jwt_auth/signer.py
dbaxa: This approach is commonly used by libraries to allow for ease of api change and not needing to keep adding arguments to the method signature.

atlassian_jwt_auth/signer.py
SamWhited: Being explicit is also future compatible unless you remove an argument (which is still trivially easy since this is only called from a single place); I don't think there's a difference except for readability.

dbaxa: I'd prefer to leave this as is as it is an internal method.

dbaxa:
Timed out.

PR #29, opened 2016-08-04
Add contrib/flask_app.py with @requires_asap to support Flask + ASAP

mark-adams:
This is based loosely off of some code that was being used in a couple
projects and I think it makes sense to institutionalize some support for
the Python web frameworks in the centralized library

- Added atlassian_jwt_auth.contrib.flask_app + tests
- Renamed atlassian_jwt_auth.contrib.requests to requests_auth to avoid
  issues with importing requests from other modules


atlassian_jwt_auth/contrib/tests/test_requests.py
```diff
@@ -5,7 +5,7 @@
 
 import atlassian_jwt_auth
 from atlassian_jwt_auth.tests import utils
-from atlassian_jwt_auth.contrib.requests import JWTAuth, create_jwt_auth
+from atlassian_jwt_auth.contrib.requests_auth import JWTAuth, create_jwt_auth
```
dbaxa: This would be breaking api change is there any reason you cannot use `from atlassian_jwt_auth.contrib.requests as requests_auth` ?

atlassian_jwt_auth/contrib/flask_app/utils.py
```diff
@@ -0,0 +1,30 @@
+from flask import current_app
+import jwt.exceptions
+import requests.exceptions
+
+
+def parse_jwt(verifier, encoded_jwt):
+    """Decode an encoded JWT using stored config."""
```
dbaxa: Personally as a consumer I'd want to know why `False` was returned especially if it was caused by an exception.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+        verifier = _get_verifier()
+        auth = request.headers.get('Authorization')
+        if not auth or not parse_jwt(verifier, auth.split(" ")[1]):
+            return Response('Unauthorized', 401)
+        return f(*args, **kwargs)
+    return decorated
+
+
+def _get_verifier():
+    """Returns a verifier based on config.asap['PUBLICKEY_REPOSITORY']"""
```
dbaxa: I don't think `PUBLICKEY_REPOSITORY` should be hard-coded here.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+
+def requires_asap(f):
+    """
+    Wrapper for Flask endpoints to make them require authentication to access.
+    """
+
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        verifier = _get_verifier()
+        auth = request.headers.get('Authorization')
```
dbaxa: Does this need to be `request.headers.get('Authorization', None)` ?

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -0,0 +1,30 @@
+from functools import wraps
+
+from flask import Response, current_app, request
+
+import atlassian_jwt_auth
+from atlassian_jwt_auth.contrib.flask_app.utils import parse_jwt
```
dbaxa: You could use a `.` relative import instead here instead
`from .utils import parse_jwt`

dbaxa:
What's the convention for flask extension / code (is `flask_app` an appropriate name for the flask contrib module) ?

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+def requires_asap(f):
+    """
+    Wrapper for Flask endpoints to make them require authentication to access.
+    """
+
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        verifier = _get_verifier()
+        auth = request.headers.get('Authorization')
+        if not auth or not parse_jwt(verifier, auth.split(" ")[1]):
```
dbaxa: I'd be careful around `auth.split(" ")[1]`  <- what if the header has an empty value (can we hit an IndexError) ?

atlassian_jwt_auth/contrib/flask_app/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        claims = verifier.verify_jwt(
+            encoded_jwt,
+            current_app.config.asap['VALID_AUDIENCE'],
+            leeway=60,
+            verify=current_app.config.asap.get('VERIFY_TLS_CERT', True)
```
dbaxa: Can you give an example when `VERIFY_TLS_CERT` would be `False`?

atlassian_jwt_auth/contrib/tests/test_flask.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def create_token(issuer, audience, key_id, private_key):
+    signer = atlassian_jwt_auth.create_signer(
+        issuer, key_id, private_key
+    )
+    return signer.generate_jwt(audience)
+
+
+class FlaskTests(utils.RS256KeyTestMixin, unittest.TestCase):
```
dbaxa: I am not accepting tests that do not have test cases that cover using the `RS256KeyTestMixin` and the `ES256KeyTestMixin` any more. Feel free to have a base test class and sub-classes that use the various mixins.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -0,0 +1,30 @@
...
+
+from flask import Response, current_app, request
+
+import atlassian_jwt_auth
+from atlassian_jwt_auth.contrib.flask_app.utils import parse_jwt
+
+
+def requires_asap(f):
+    """
+    Wrapper for Flask endpoints to make them require authentication to access.
```
dbaxa: I suggest changing `Wrapper for Flask endpoints to make them require authentication to access.` to `Wrapper for Flask endpoints to make them require asap authentication to access.`

test-requirements.txt
```diff
@@ -1,2 +1,3 @@
 nose
 mock
+flask
```
dbaxa: Please specify a version range for flask. The tests shouldn't fail if flask releases a new major version that causes our flask related tests to fail.

atlassian_jwt_auth/contrib/tests/test_requests.py
mark-adams: Yea, so this was a major mistake on my part originally. Having the module named `requests` blocks sibling modules from using `import requests` to import the real `requests` module.

It is a breaking API change but I think its an intelligent one.

atlassian_jwt_auth/contrib/flask_app/utils.py
mark-adams: Yea, I know that I submitted this code... but this is a direct copy & paste from the `asap.py` file in Instant Micros :cry: 

Not saying that I'm opposed to changing it, but just establishing why this is the way it is

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: I'm not sure I understand your opposition to the hardcoding. `PUBLICKEY_REPOSITORY` is namespaced within the `asap` section of the `config`. That seems equivalent to your `ASAP_PUBLICKEY_REPOSITORIES` sugguestion

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: No, that's one of the differences between `request.headers['Authorization']` and `request.headers.get('Authorization')`. The latter will return `None`. The former will raise a `KeyError`

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: Great suggestion!

- [ ] Change the import to be relatie

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: Yep. Sure can. This was one of the copy & paste's from Instant Micros.

- [ ] Make auth header parsing more robust

atlassian_jwt_auth/contrib/flask_app/utils.py
mark-adams: Obviously the token should be a required parameter. Valid audience seems like something we definitely want (I can't think of a reason why an ASAP service wouldn't want to validate that a token wasn't being re-used across services). It might make sense to customize leeway.

Yet again, `VERIFY_TLS_CERT` was part of the original copy & paste from the Instant Micros template. I think the main reasoning in its existence was for testing situtations. Its a significant amount of extra work to spin up a full TLS server to simulate a key repository but with my testing utils class that I added, we probably don't need this at all anymore. I'd be open to removing it

atlassian_jwt_auth/contrib/tests/test_flask.py
mark-adams: I'm not sure why that's necessary. We're extensively testing the underlying implementation directly. I think it only adds brittleness to the test suite to focus on that in here.

In other words, these tests should focus specifically on the behavior of the decorator itself under different scenarios; not the behavior of the underlying verifier.

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: Sounds great!

- [] Update docstring for `@require_asap`

test-requirements.txt
mark-adams: Maybe? I've found that to actually be useful in a number of scenarios to help detect when Flask makes breaking changes so that we can address those incompatibilities.

In many projects, I've seen the dep left without a version range and then, if the case arises where the tests are failing because of a new flask version causing breakage and its blocking something else from going out, we can add the range at that time.

But I don't feel really strongly about it; just sharing my experience from a couple other Python projects.

atlassian_jwt_auth/contrib/flask_app/__init__.py
```diff
@@ -0,0 +1,2 @@
+
```
mark-adams: @dbaxa So, the main issue that this naming strategy is trying to fix is the clobbering of core packages like `flask`, `django`, and `requests` that prevent sibling packages under `contrib` from using those libraries.

Not many libraries have this problem where they include stubbing for Django, Flask, etc, but the most notable, `python-social-auth` which contains the "Login with X" code for almost every cloud service and it follows the pattern of `django_app`, `flask_app`, etc. as you see [here](https://github.com/omab/python-social-auth/tree/master/social/apps). I think its a good and valuable pattern in this scenario.

atlassian_jwt_auth/contrib/tests/test_requests.py
dbaxa: I am not sold on this change. Can we drop it from this pull request :-) ?

atlassian_jwt_auth/contrib/flask_app/utils.py
dbaxa: I'd rather leave catching the errors to callers and note what they should be (exceptions).

atlassian_jwt_auth/contrib/flask_app/decorators.py
dbaxa: Ah okay.

atlassian_jwt_auth/contrib/flask_app/utils.py
dbaxa: Let's remove it and people who need to work with TLS configurations where there isn't a trusted certificate in use will have to write some more code.

test-requirements.txt
dbaxa: I do feel strongly about pinning an upper bound. I don't want the tests to break because of flask code changes.

atlassian_jwt_auth/contrib/flask_app/__init__.py
dbaxa: Okay.

atlassian_jwt_auth/contrib/flask_app/utils.py
mark-adams: I'm going to push back on this one. I don't see any reason to make a developer spin up generate a TLS server and start a web server just to run a test. Its very common to make this available to the user for this purpose.

What about compromising by emitting a warning whenever `VERIFY_TLS_CERT` is `False` and `debug` is not `True`?

dbaxa: @mark-adams nah. I'd rather not compromise.

atlassian_jwt_auth/contrib/flask_app/utils.py
```diff
@@ -0,0 +1,18 @@
...
+        encoded_jwt,
+        current_app.config.asap['VALID_AUDIENCE'],
+        leeway=60,
+        verify=current_app.config.asap.get('VERIFY_TLS_CERT', True)
+    )
+
+    valid_issuers = current_app.config.asap.get('VALID_ISSUERS')
+    if valid_issuers and claims.get('iss') not in valid_issuers:
+        raise InvalidIssuerError
+
```
dbaxa: What's with all the unneeded newlines ?

atlassian_jwt_auth/contrib/flask_app/utils.py
```diff
@@ -0,0 +1,18 @@
...
+from jwt.exceptions import InvalidIssuerError
+
+
+def parse_jwt(verifier, encoded_jwt):
+    """Decode an encoded JWT using stored config."""
+    claims = verifier.verify_jwt(
+        encoded_jwt,
+        current_app.config.asap['VALID_AUDIENCE'],
+        leeway=60,
+        verify=current_app.config.asap.get('VERIFY_TLS_CERT', True)
```
dbaxa: I still feel that I want this removed.

atlassian_jwt_auth/contrib/flask_app/utils.py
mark-adams: I don't see any unneeded newlines?

atlassian_jwt_auth/contrib/flask_app/utils.py
mark-adams: :+1: Removed

PR #30, opened 2016-08-09
WIP: Initial implementation of Django support

mark-adams:
This is still in-progress. It is mostly complete; just needs tests.


atlassian_jwt_auth/contrib/django_app/migrations/0001_initial.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10 on 2016-08-09 02:33
```
dbaxa: We probably don't need these `generated` lines.

atlassian_jwt_auth/contrib/flask_app/utils.py
```diff
@@ -1,16 +1,12 @@
-from flask import current_app
 from jwt.exceptions import InvalidIssuerError
 
 
-def parse_jwt(verifier, encoded_jwt):
+def parse_jwt(verifier, encoded_jwt, valid_audience, valid_issuers):
```
dbaxa: This is kind of a breaking change ... perhaps we should go with `kwargs` ?

mark-adams: I removed it

atlassian_jwt_auth/contrib/django_app/auth.py
```diff
@@ -0,0 +1,44 @@
...
+        except (
+            requests.exceptions.HTTPError,
+            requests.exceptions.ConnectionError,
+            jwt.Exceptions.InvalidTokenError
+        ):
+            logging.exception('Authentication failed while verifying ASAP token')
+            return None
+
+        try:
+            return Issuer.objects.get(issuer=claims['iss']).user
```
dbaxa: `.user` ?

atlassian_jwt_auth/contrib/django_app/auth.py
```diff
@@ -0,0 +1,44 @@
...
+        except Issuer.DoesNotExist:
+            return None
+
+
+def _get_verifier():
+    """Returns a verifier based on config.asap['PUBLICKEY_REPOSITORY']"""
+    return atlassian_jwt_auth.JWTAuthVerifier(
+        atlassian_jwt_auth.HTTPSPublicKeyRetriever(
+            settings.ASAP_PUBLICKEY_REPOSITORY
```
dbaxa: Please use `getattr(settings, 'ASAP_PUBLICKEY_REPOSITORY', None)` as `settings.ASAP_PUBLICKEY_REPOSITORY` may not be set.

atlassian_jwt_auth/contrib/django_app/decorators.py
```diff
@@ -0,0 +1,41 @@
+from functools import wraps
+from django.http import HttpResponse, HttpResponseForbidden
+from django.conf import settings
+
+import jwt
+
+from atlassian_jwt_auth.contrib.django_app.auth import ASAPAuthBackend
+
+
+class HttpResponseNotAuthorized(HttpResponse):
```
dbaxa: I wonder why django doesn't provide this for us ...

atlassian_jwt_auth/contrib/django_app/decorators.py
```diff
@@ -0,0 +1,41 @@
...
+    status_code = 401
+
+
+def asap_required(view_func, issuers=None):
+    """Decorator which ensures the user has provided a correct user and token pair."""
+
+    @wraps(view_func)
+    def _wrapped_view(request, *args, **kwargs):
+        user = request.user
+
```
dbaxa: We don't need so many blank lines in this method.

atlassian_jwt_auth/contrib/django_app/decorators.py
```diff
@@ -0,0 +1,41 @@
...
+            return HttpResponseNotAuthorized('This endpoint requires ASAP authentication.')
+
+        claims = jwt.decode(token, verify=False)
+        if issuers and claims.get('iss') not in issuers:
+            return HttpResponseForbidden('ASAP issuer is not allowed to access this endpoint')
+
+        if not issuers and settings.ASAP_ALLOWED_ISSUERS and claims.get('iss') not in issuers:
+            return HttpResponseForbidden('ASAP issuer is not allowed to access this endpoint')
+
+        request.user = user
```
dbaxa: Are we sure we want to map it to a user ?

examples/django_example/django_example/settings.py
```diff
@@ -0,0 +1,121 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/1.10/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = 'p&e^acy-(!!@ubdlj6-ncob^^#gvzq=tjfz7qvm5kj*2prddwz'
```
dbaxa: IMHO in the example settings we should generate this randomly (every time).

atlassian_jwt_auth/contrib/django_app/auth.py
mark-adams: `Issuer.user` represents a specific Django user.

In Django, all authentication backends are required to take some sort of credential and return a matching user. The best way to do this in the context of ASAP is to validate the token, extract the issuer, and then look up the `iss` value and determine if it matches a user. If Django finds a user linked to that `iss` (via `Issuer`), the user is authenticated. If Django does not find a user, authentication fails even though the token is valid (since the service doesn't recognize the issuer)

atlassian_jwt_auth/contrib/django_app/decorators.py
mark-adams: I did find that peculiar as well. I suppose they figure that generally someone writing a Django application and using stock backends will not need to raise a 401 since the framework will do it for them?

atlassian_jwt_auth/contrib/django_app/decorators.py
mark-adams: In this case, the blank lines improve readability of the function and provide clear separation between the docstring and the nested function. This is not an abnormal styling in most Python projects.

atlassian_jwt_auth/contrib/django_app/decorators.py
mark-adams: Yes, Django's authentication model requires authentication backends tie credentials to a specific user.

atlassian_jwt_auth/contrib/django_app/apps.py
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class ASAPAuthConfig(AppConfig):
```
dbaxa: docstring is appreciated :-)

mark-adams:
Declining this in favor of #35

PR #32, opened 2016-08-09
Add better settings for test

mark-adams:
This PR makes testing with Flask MUCH easier by adding an optional ASAP_KEY_RETRIEVER_CLASS setting to specify a class that can be used instead of the default for retrieving public keys.

### DIFF
--- a/atlassian_jwt_auth/contrib/flask_app/decorators.py
+++ b/atlassian_jwt_auth/contrib/flask_app/decorators.py
@@ -46,9 +46,7 @@ def _get_verifier():
     retriever_cls = current_app.config.get(
         'ASAP_KEY_RETRIEVER_CLASS', atlassian_jwt_auth.HTTPSPublicKeyRetriever
     )
-
     retriever = retriever_cls(
         base_url=current_app.config.get('ASAP_PUBLICKEY_REPOSITORY')
     )
-
     return atlassian_jwt_auth.JWTAuthVerifier(retriever)
--- a/atlassian_jwt_auth/contrib/flask_app/decorators.py
+++ b/atlassian_jwt_auth/contrib/flask_app/decorators.py
@@ -48,7 +48,7 @@ def _get_verifier():
     )
 
     retriever = retriever_cls(
-        base_url=current_app.config['ASAP_PUBLICKEY_REPOSITORY']
+        base_url=current_app.config.get('ASAP_PUBLICKEY_REPOSITORY')
     )
 
     return atlassian_jwt_auth.JWTAuthVerifier(retriever)
--- a/atlassian_jwt_auth/contrib/flask_app/decorators.py
+++ b/atlassian_jwt_auth/contrib/flask_app/decorators.py
@@ -42,9 +42,13 @@ def requires_asap(f):
 
 
 def _get_verifier():
-    """Returns a verifier based on config['ASAP_PUBLICKEY_REPOSITORY']"""
-    return atlassian_jwt_auth.JWTAuthVerifier(
-        atlassian_jwt_auth.HTTPSPublicKeyRetriever(
-            current_app.config['ASAP_PUBLICKEY_REPOSITORY']
-        )
+    """Returns a verifier for ASAP JWT tokens basd on application settings"""
+    retriever_cls = current_app.config.get(
+        'ASAP_KEY_RETRIEVER_CLASS', atlassian_jwt_auth.HTTPSPublicKeyRetriever
     )
+
+    retriever = retriever_cls(
+        base_url=current_app.config['ASAP_PUBLICKEY_REPOSITORY']
+    )
+
+    return atlassian_jwt_auth.JWTAuthVerifier(retriever)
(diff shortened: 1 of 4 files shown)


atlassian_jwt_auth/contrib/tests/utils.py
```diff
@@ -1,20 +1,21 @@
 import atlassian_jwt_auth
 
 
-class StaticPublicKeyRetriever(object):
```
dbaxa: What value does changing this code add? (I don't see any value except code churn ...).

mark-adams: I'm glad you asked. The main point of this PR is to expose a setting that is obvious and easy to access during test to allow tests to substitute a public key retriever during tests. Previously, this was very challenging without using `mock.patch()` on some internal implementation detail of the library.

The only problem with my solution is that it expects the user to provide a concrete class that can be instantiated by the Flask plugin to serve as the public key retriever. I wanted to be able to pass keys into that class though and the best way to do that was to create a factory function that would create an anonymous class that contained the static keys that I wanted to use.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -42,9 +42,13 @@ def decorated(*args, **kwargs):
 
 
 def _get_verifier():
-    """Returns a verifier based on config['ASAP_PUBLICKEY_REPOSITORY']"""
-    return atlassian_jwt_auth.JWTAuthVerifier(
-        atlassian_jwt_auth.HTTPSPublicKeyRetriever(
-            current_app.config['ASAP_PUBLICKEY_REPOSITORY']
-        )
+    """Returns a verifier for ASAP JWT tokens basd on application settings"""
+    retriever_cls = current_app.config.get(
```
dbaxa: Can we remove most if not all o the empty lines from this method?

atlassian_jwt_auth/contrib/tests/utils.py
dbaxa: I am not disagreeing with that part. I am asking why we need this code churn. You need not move the `StaticPublicKeyRetriever` class under the `get_static_retriever_class` class and in fact I personally disagree with doing so.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -42,9 +42,13 @@ def decorated(*args, **kwargs):
...
+        'ASAP_KEY_RETRIEVER_CLASS', atlassian_jwt_auth.HTTPSPublicKeyRetriever
     )
+
+    retriever = retriever_cls(
+        base_url=current_app.config['ASAP_PUBLICKEY_REPOSITORY']
```
dbaxa: If `ASAP_PUBLICKEY_REPOSITORY` is not configured what happens here?

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: I think the formatting here is fine and consistent with most of our Python coding standards used by Atlassian teams. These spaces improve readability of the code and seem proper to me.

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: A `KeyError` is raised

Fixed!

atlassian_jwt_auth/contrib/tests/utils.py
dbaxa: Let me think about it.

atlassian_jwt_auth/contrib/flask_app/decorators.py
dbaxa: For such a short method I feel that the newlines do not help with readability at all and dislike them here.

atlassian_jwt_auth/contrib/tests/utils.py
mark-adams: OK :smile:

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: I have removed the spaces as requested

atlassian_jwt_auth/key.py
```diff
@@ -59,7 +59,7 @@ class HTTPSPublicKeyRetriever(object):
     """
 
     def __init__(self, base_url):
-        if not base_url.startswith('https://'):
+        if base_url is None or not base_url.startswith('https://'):
```
dbaxa: I'd suggest leaving this change for another pull request (as it is unrelated).

atlassian_jwt_auth/contrib/tests/utils.py
dbaxa: I think this is being a bit overly clever and that a simpler and better solution is to have the flask `requires_asap` decorator work with a `verifier` keyword argument - does that make sense? (my memory of decorators is a little bit rusty)

mark-adams: The main problem with the `@requires_asap` is that you can't modify the decorator at runtime, only a setting  allows for that.

PR #35, opened 2016-12-25
BBCDEV-4046 Add Django support

digimarc:
This is pretty much the same as the Flask decorator, but Django flavored. One key difference is that the Django decorator requires an issuer. The allows/ensures that specific views are only accessible by specific issuers.


atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+    """Decorator for Django endpoints to require ASAP
+
+    :param list issuers: (required) The 'iss' claims that this endpoint is from
+    """
+    def requires_asap_decorator(func):
+        @wraps(func)
+        def requires_asap_wrapper(request, *args, **kwargs):
+            verifier = _get_verifier()
+            auth = request.META.get('AUTHORIZATION', '').split(' ')
+            print(request.META, auth)
```
dbaxa: Remove this print.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+
+        return requires_asap_wrapper
+    return requires_asap_decorator
+
+
+def _get_verifier():
+    """Return a verifier for ASAP JWT tokens based on settings"""
+    retriever_cls = getattr(settings, 'ASAP_KEY_RETRIEVER_CLASS',
+                            atlassian_jwt_auth.HTTPSPublicKeyRetriever)
+
```
dbaxa: Please remove the unneeded whitespace.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def parse_jwt(verifier, encoded_jwt):
+    """Decode an encoded JWT using stored config."""
+    claims = verifier.verify_jwt(
+        encoded_jwt,
+        settings.ASAP_VALID_AUDIENCE,
+        leeway=60
+    )
+
```
dbaxa: Please remove the unneeded whitespace.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    return claims
+
+
+def verify_issuers(asap_claims, issuers):
+    """Verify that the issuer in the claims is valid and is expected"""
```
dbaxa: The docstring in the previous method ends with a `.` so to be consistent this one should also end with a `.`.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    if asap_claims.get('iss') not in issuers:
+        # Raise early if the specific issuer isn't expected
+        raise InvalidIssuerError()
+
+    valid_issuers = settings.ASAP_VALID_ISSUERS
```
dbaxa: We are not using `getattr` here because we do not want this to "fall open" if an application lacks a `ASAP_VALID_ISSUERS` setting?

atlassian_jwt_auth/contrib/tests/django/settings.py
```diff
@@ -0,0 +1,128 @@
+"""
+Django settings for foop project.
```
dbaxa: This docstring isn't needed.

atlassian_jwt_auth/contrib/tests/django/settings.py
```diff
@@ -0,0 +1,128 @@
...
+                'django.template.context_processors.debug',
+                'django.template.context_processors.request',
+                'django.contrib.auth.context_processors.auth',
+                'django.contrib.messages.context_processors.messages',
+            ],
+        },
+    },
+]
+
+# WSGI_APPLICATION = 'foop.wsgi.application'
```
dbaxa: You can remove this if this is not needed.

atlassian_jwt_auth/contrib/tests/django/settings.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+# Database
+# https://docs.djangoproject.com/en/1.10/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.sqlite3',
+        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
```
dbaxa: Apparently for the test sqlite database you can set `NAME` as `None` to use an in-memory database.

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,85 @@
+import django
```
dbaxa: This `django` import shouldn't be here - it should be with the django imports bellow (it isn't a standard module).

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,85 @@
...
+
+from django.test.testcases import SimpleTestCase
+from django.test.utils import override_settings
+
+from atlassian_jwt_auth import create_signer
+from atlassian_jwt_auth.contrib.tests.utils import get_static_retriever_class
+from atlassian_jwt_auth.tests import utils
+
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE',
```
dbaxa: I feel that you may want this to go into a setup / tear down method (that isn't required but it might be preferable).

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def create_token(issuer, audience, key_id, private_key):
+    signer = create_signer(issuer, key_id, private_key)
+
+    return signer.generate_jwt(audience)
+
+
+class TestAsapDecorator(SimpleTestCase):
+    _private_key_pem = utils.get_new_rsa_private_key_in_pem_format()
```
dbaxa: I am requiring that new tests use the test mixins that are available so that you can test both RSA and ECDSA algorithms - please update the code accordingly.

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
 nose
 mock
 flask<0.12
+django<1.11
```
dbaxa: Hmm. I think specifying a lower bound to be specific about what Django versions the contrib module "supports" might be useful.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+                return HttpResponse('Unauthorized', status=401)
+
+            try:
+                asap_claims = parse_jwt(verifier, auth[1])
+                verify_issuers(asap_claims, issuers)
+                request.asap_claims = asap_claims
+                return func(request, *args, **kwargs)
+            except HTTPError:
+                # Couldn't find key in key server
+                return HttpResponse('Unauthorized: Invalid key', status=401)
```
dbaxa: This is consistent with the flask code.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+            except ConnectionError:
+                # Also couldn't find key in key-server
+                return HttpResponse(
+                    'Unauthorized: Backend server connection error',
```
dbaxa: Logging what happened here could be useful - although the flask implementation doesn't log at the present time either.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+            print(request.META, auth)
+            if not auth or len(auth) != 2:
+                return HttpResponse('Unauthorized', status=401)
+
+            try:
+                asap_claims = parse_jwt(verifier, auth[1])
+                verify_issuers(asap_claims, issuers)
+                request.asap_claims = asap_claims
+                return func(request, *args, **kwargs)
+            except HTTPError:
```
dbaxa: I suggest setting an error message based upon the exception type and then having a check to see if the error message has been set and return a HttpResponse like so ->

```
if error:
       return HttpResponse(error, status=401)
```

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+from requests.exceptions import (HTTPError, ConnectionError)
+from jwt.exceptions import (InvalidIssuerError, InvalidTokenError)
+
+from atlassian_jwt_auth.contrib.django.utils import parse_jwt, verify_issuers
+
+
+def requires_asap(issuers):
+    """Decorator for Django endpoints to require ASAP
+
+    :param list issuers: (required) The 'iss' claims that this endpoint is from
```
dbaxa: Missing `.`.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    def requires_asap_decorator(func):
+        @wraps(func)
+        def requires_asap_wrapper(request, *args, **kwargs):
+            verifier = _get_verifier()
```
dbaxa: I haven't checked how if creating a new `verifier` like this has an impact on cachecontrol (and caching of public keys) but you may want to check.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    claims = verifier.verify_jwt(
+        encoded_jwt,
+        settings.ASAP_VALID_AUDIENCE,
+        leeway=60
+    )
+
+    return claims
+
+
+def verify_issuers(asap_claims, issuers):
```
dbaxa: I'd actually call this something like `verify_jwt_issuer` and have it take `token` or `a_jwt` & `allowed_issuers`.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    """Verify that the issuer in the claims is valid and is expected"""
+    if asap_claims.get('iss') not in issuers:
```
dbaxa: The flask code allows `issuers` to be an empty set or an empty list, this code should allow the same.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,26 @@
...
+        # Raise early if the specific issuer isn't expected
+        raise InvalidIssuerError()
+
+    valid_issuers = settings.ASAP_VALID_ISSUERS
+    if valid_issuers and asap_claims.get('iss') not in valid_issuers:
+        raise InvalidIssuerError()
+
+    return
```
dbaxa: We don't need this explicit return.

test-requirements.txt
mark-adams: Yea, I'd like to see this changed to `Django>=1.8` since 1.8 is the oldest supported Django.

I'm working on a PR to add a tox build that will run the tests against the Python 2/3 and Django version combinations.

atlassian_jwt_auth/contrib/django/utils.py
digimarc: Correct. If you choose to use this library, this is required. That was the idea anyway. I could also default to an empty list if that is preferred.

atlassian_jwt_auth/contrib/django/utils.py
digimarc: I disagree. I think it is important that each endpoint/view specify *which* issuers are allowed to access it. Doing this ensures that it is more secure and narrow by default.

atlassian_jwt_auth/contrib/django/decorators.py
digimarc: > 79 characters :(

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,142 @@
...
+
+from atlassian_jwt_auth import create_signer
+from atlassian_jwt_auth.contrib.tests.utils import get_static_retriever_class
+from atlassian_jwt_auth.tests import utils
+from atlassian_jwt_auth.tests.utils import RS256KeyTestMixin
+
+
+def create_token(issuer, audience, key_id, private_key):
+    signer = create_signer(issuer, key_id, private_key)
+
```
dbaxa: This whitespace isn't needed.

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+    def test_request_with_valid_token_is_allowed(self):
+        token = create_token(
+            issuer='client-app', audience='server-app',
+            key_id='client-app/key01', private_key=self._private_key_pem
+        )
+
+        with override_settings(**self.test_settings):
+            response = self.client.get('/asap/test1',
```
dbaxa: You could use django's URL reverse url lookup code here.

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
 nose
 mock
 flask<0.12
+django>=1.8
```
dbaxa: Let's go with Django>=1.8,<1.11

atlassian_jwt_auth/contrib/tests/django/test_django.py
digimarc: Is this a deviation from the pep8 recommendation? https://www.python.org/dev/peps/pep-0008/#blank-lines

test-requirements.txt
digimarc: (nods)

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -0,0 +1,22 @@
...
+        encoded_jwt,
+        settings.ASAP_VALID_AUDIENCE,
+        leeway=60
+    )
+    return claims
+
+
+def verify_issuers(asap_claims, issuers):
+    """Verify that the issuer in the claims is valid and is expected."""
+    if asap_claims.get('iss') not in issuers:
```
dbaxa: @digimarc  so do we want issuers to never be empty ?

dbaxa:
When merging this please squash the commits and use a commit message similar to the following:

```Sem-Ver: feature Added Django support.
* Add contrib/django/decorators.py with @requires_asap to support Django + ASAP.
```

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class TestAsapDecorator(RS256KeyTestMixin, SimpleTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        os.environ.setdefault(
+            'DJANGO_SETTINGS_MODULE',
+            'atlassian_jwt_auth.contrib.tests.django.settings')
+        django.setup()
```
mark-adams: Do you really have to do django.setup() if you're running inside a Django test case?

atlassian_jwt_auth/contrib/tests/django/views.py
```diff
@@ -0,0 +1,23 @@
...
+def unexpected_view(request):
+    return HttpResponse('This should fail.')
+
+
+@requires_asap(issuers=['whitelist'])
+def decorated_view(request):
+    return HttpResponse('Only the right issuer is allowed.')
+
+
+@requires_asap()
```
mark-adams: You don't need the parens here

PR #36, opened 2017-01-12
Allow to pass in additional claims to contrib.requests.JWTAuth

whummer:
This PR contains a minor change to allow passing in additional claims via `**kwargs` to `contrib.requests.JWTAuth`. As far as I can tell, there is currently no other way to do this, unless I'm missing something. /cc @dbaxa

### DIFF
--- a/atlassian_jwt_auth/contrib/requests.py
+++ b/atlassian_jwt_auth/contrib/requests.py
@@ -17,13 +17,15 @@ class JWTAuth(AuthBase):
     """Adds a JWT bearer token to the request per the ASAP specification"""
 
     def __init__(self, signer, audience, *args, **kwargs):
-        super(JWTAuth, self).__init__(*args, **kwargs)
+        super(JWTAuth, self).__init__()
 
         self._audience = audience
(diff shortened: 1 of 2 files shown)


whummer:
Tests are currently failing. Looks like the installed version of `jinja2` uses some python 3 features (`async def`) which result in syntax errors. We might need to adjust the version in `requirements.txt`

atlassian_jwt_auth/contrib/requests.py
```diff
@@ -17,13 +17,16 @@ class JWTAuth(AuthBase):
     """Adds a JWT bearer token to the request per the ASAP specification"""
 
     def __init__(self, signer, audience, *args, **kwargs):
-        super(JWTAuth, self).__init__(*args, **kwargs)
+        # Note: requests.auch.AuthBase.__init__() does not accept parameters
```
dbaxa: I feel that this comment isn't needed. Can you remove it?

whummer: Done.

PR #41, opened 2017-04-20
Warn when aiohttp is not installed so that the tests do not fail in python >= 3.5 when aiohttp is not installed.

dbaxa:


dbaxa:
cc @Fedorof

atlassian_jwt_auth/contrib/aiohttp/__init__.py
```diff
@@ -2,8 +2,15 @@
...
+    try:
+        import aiohttp
+        from .auth import JWTAuth
+        from .key import HTTPSPublicKeyRetriever
+        from .verifier import JWTAuthVerifier
+    except ImportError:
+        import warnings
+        warnings.warn(
+            'Could not import aiohttp code as aiohttp is not installed.')
+
```
Fedorof: `import aiohttp` theoretically may raise `ImportError` when one of its dependencies is not installed. 

`from .auth import JWTAuth` and other imports below it may raise ImportError due to a bug, or when some dependencies are not installed as well.

So, I would do this:

```
    . . .
except ImportError as err:
    warnings.warn(str(err))
```

dbaxa: Okay will do.

PR #42, opened 2017-06-27
Fix Python 3 issue with default value

codingkoi:
When using the default value in the `requires_asap` decorator, the
`split(b'')` call fails in Python 3 because bytes != str. This change just
sets the default value to a bytes instead of a str.


mark-adams:
That looks good to me. @dbaxa Are you alright with me merging this?

PR #44, opened 2017-06-29
Fix for using requests library in Python 3

jeremyshoemaker:
More issues with string encodings. The requests library seems to force
HTTP headers to be strings even if you give it a bytes so to deal with
that potential ambiguity, we should just explicitly convert any
strings we get in the HTTP_AUTHORIZATION header into bytes before
attempting to parse it further.

### DIFF
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -2,8 +2,9 @@ from functools import wraps
 
 from django.conf import settings
 from django.http.response import HttpResponse
-from requests.exceptions import (HTTPError, ConnectionError)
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -1,5 +1,6 @@
 from functools import wraps
 
+import six
```
dbaxa: Why are you using `six` explicitly here and not django's `six` ?

jeremyshoemaker: That's fine. I haven't had to add it to a project before, it's usually already present, and I forgot about Django having a copy.

PR #45, opened 2017-07-05
Sem-Ver: feature Add support for customising the value of the leeway …

dbaxa:
…used in the django and flask contrib code through the ASAP_VALID_LEEWAY setting.

Please note as part of this change the default leeway has been changed to 0 for consistency.

### DIFF
--- a/atlassian_jwt_auth/contrib/django/utils.py
+++ b/atlassian_jwt_auth/contrib/django/utils.py
@@ -7,7 +7,7 @@ def parse_jwt(verifier, encoded_jwt):
     claims = verifier.verify_jwt(
         encoded_jwt,
         settings.ASAP_VALID_AUDIENCE,
-        leeway=60
+        leeway=getattr(settings, 'ASAP_VALID_LEEWAY', 0)
     )
     return claims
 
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -7,7 +7,7 @@ def parse_jwt(verifier, encoded_jwt):
     claims = verifier.verify_jwt(
         encoded_jwt,
         settings.ASAP_VALID_AUDIENCE,
-        leeway=60
+        leeway=getattr(settings, 'ASAP_VALID_LEEWAY', 0)
```
dbaxa: IMHO 60 seconds is too long when that's the suggested token lifetime. Allowing users to specify their own value and defaulting to zero seems fairly sane to me (given the short token lifetime) do you have issues with time synchronisation between severs that would require a value higher than say `5` here?

cc @mark-adams / @jeremyshoemaker

dbaxa: Also I am happy to make this an api-break if you feel that's a good idea.

jeremyshoemaker: Seems like you're breaking the API here, so if that's the way you want to go with it, then api-break makes sense to me. That said, I'm not sure that a TTL of 0 seconds would ever make sense. Wouldn't it expire the moment it's created?

If I were doing it, I would probably add overriding the value as a feature and leave the default at 60 then later make an api-break to change the default if the need arises. I guess I'm not a big fan of breaking APIs.

dbaxa: I have gone with a default leeway of 0 seconds.

PR #46, opened 2017-07-05
Make _seen_jti a ringbuffer.

erikvanzijst:
Throwing away the entire contents of the jti list when it reaches capacity seems undesirable. This change makes it a ringbuffer that automatically takes care of expunging old entries.

I've also increased the capacity.

### DIFF
--- a/atlassian_jwt_auth/verifier.py
+++ b/atlassian_jwt_auth/verifier.py
@@ -1,4 +1,4 @@
-from collections import deque, OrderedDict
+from collections import OrderedDict
 
 import jwt
 
--- a/atlassian_jwt_auth/verifier.py
+++ b/atlassian_jwt_auth/verifier.py
@@ -1,3 +1,5 @@
+from collections import deque
+
 import jwt
 
 from atlassian_jwt_auth import algorithms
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/verifier.py
```diff
@@ -11,7 +13,7 @@ class JWTAuthVerifier(object):
     def __init__(self, public_key_retriever, **kwargs):
         self.public_key_retriever = public_key_retriever
         self.algorithms = algorithms.get_permitted_algorithm_names()
-        self._seen_jti = set()
+        self._seen_jti = deque(maxlen=1000)
```
dbaxa: I don't mind having a backing deque like object but the (average) time complexity is for searching for a jti is higher for `deque` than it is for a `set` (O(n) vs O(1)).

erikvanzijst: That's fair. I'll switch it to log(n).

atlassian_jwt_auth/verifier.py
```diff
@@ -1,3 +1,5 @@
+from collections import deque, OrderedDict
```
dbaxa: We don't need the `deque` import any longer.

dbaxa:
Will re-raise.

PR #47, opened 2017-07-25
Add multiple public key retriever support to library

mbertrand80:
Thanks to BJacobs for the base work here. I took it the rest of the way to make sure it works with Python2-3 as well as make some of the exception raising and handling a bit more pythonic. I also added a lot more information and logging to the decorator for Django. This required changing some expected test output.

### DIFF
--- a/atlassian_jwt_auth/contrib/tests/django/test_django.py
+++ b/atlassian_jwt_auth/contrib/tests/django/test_django.py
@@ -88,7 +88,7 @@ class TestAsapDecorator(RS256KeyTestMixin, SimpleTestCase):
             response = self.client.get(reverse('expected'),
                                        HTTP_AUTHORIZATION=b'Bearer ' + token)
 
-        self.assertContains(response, 'Unauthorized: Invalid token',
+        self.assertContains(response, 'Unauthorized: Invalid audience',
                             status_code=401)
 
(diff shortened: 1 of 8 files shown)


atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -119,7 +119,9 @@ def test_request_with_invalid_issuer_is_rejected(self):
             response = self.client.get(reverse('expected'),
                                        HTTP_AUTHORIZATION=b'Bearer ' + token)
 
-        self.assertContains(response, 'Unauthorized: Invalid token issuer',
+        self.assertContains(response,
```
dbaxa: I don't think we need to have the issuer as part of the response error content.

atlassian_jwt_auth/key.py
```diff
@@ -6,9 +6,14 @@
 
 import cachecontrol
 import cryptography.hazmat.backends
-from cryptography.hazmat.primitives import serialization
 import jwt
 import requests
+from cryptography.hazmat.primitives import serialization
+from requests.exceptions import RequestException
+
```
dbaxa: Why is there an blank line here?

mbertrand80:
This needs quite a bit of work.

PR #48, opened 2017-07-26
Add library exceptions instead of using ValueError

mbertrand80:
This also maintains backwards compatibility with existing clients
by keeping ValueError as the base exception.

I also added some logging for Django, but not opposed to splitting that out too.

### DIFF
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -12,8 +12,6 @@ from atlassian_jwt_auth.exceptions import (PrivateKeyRetrieverException,
                                            PublicKeyRetrieverException)
 from .utils import parse_jwt, verify_issuers
 
-logger = logging.getLogger(__name__)
-
 
 def requires_asap(issuers=None):
     """Decorator for Django endpoints to require ASAP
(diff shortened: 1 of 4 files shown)


atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -1,14 +1,19 @@
...
 from jwt.exceptions import (InvalidIssuerError, InvalidTokenError)
-from requests.exceptions import ConnectionError, HTTPError
+from requests.exceptions import RequestException
 
 import atlassian_jwt_auth
+from atlassian_jwt_auth.exceptions import (PrivateKeyRetrieverException,
+                                           PublicKeyRetrieverException)
 from .utils import parse_jwt, verify_issuers
 
+logger = logging.getLogger(__name__)
```
dbaxa: In general I am not a fan of having globals like this. Because this logger is currently only used inside the `requires_asap` method please move it inside that method.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -1,7 +1,12 @@
+import logging
+
 from django.conf import settings
 from jwt.exceptions import InvalidIssuerError
 
 
+logger = logging.getLogger(__name__)
```
dbaxa: Please move this to under `verify_issuers` because it is currently only used there.

PR #49, opened 2017-07-26
Add HTTPSMultiKeyRetriever and tests

mbertrand80:
This works exactly like HTTPSKeyRetriever except that it accepts an arbitrarily long list of keystore URLs.


atlassian_jwt_auth/key.py
```diff
@@ -1,21 +1,24 @@
...
+from cryptography.hazmat.primitives import serialization
+from requests.exceptions import RequestException
 
 if sys.version_info[0] >= 3:
     from urllib.parse import unquote_plus
 else:
     from urllib import unquote_plus
 
 PEM_FILE_TYPE = 'application/x-pem-file'
+logger = logging.getLogger(__name__)
```
dbaxa: I prefer to avoid globals such as this. Perhaps make a `_get_logger()` method  or obtain a logger where you need it.

atlassian_jwt_auth/key.py
```diff
@@ -96,6 +111,38 @@ def _check_content_type(self, url, content_type):
                              (content_type, url))
 
 
+class HTTPSMultiPublicKeyRetriever(HTTPSPublicKeyRetriever):
```
dbaxa: Can we have some docstring for this class :-) ?

mbertrand80: Done

dbaxa:
This feature was added by #50. You may want to open another pr for the django change @mbertrand80  :-)

PR #50, opened 2017-07-28
Sem-Ver: feature Add a new HTTPSMultiRepositoryPublicKeyRetriever class which allows using multiple public key repositories.

dbaxa:


dbaxa:
cc @mbertrand80

PR #55, opened 2017-09-27
Add Django middleware to auth forwarded clients

grantisu:
Some services will proxy client requests to other services, where client
does not use ASAP and the proxy must rely on other, ASAP-enabled
services to authenticate the client.

To support this, we allow the proxy to add the client's authorization
header as X-ASAP-Authorization. The middleware will then swap
authorization headers before any authorization middleware sees the
request, and swap the headers back before any view-level ASAP checks are
done.

To support anonymous clients, we require a second X-ASAP header before
we swap authorization headers.

### DIFF
--- a/atlassian_jwt_auth/contrib/django/middleware.py
+++ b/atlassian_jwt_auth/contrib/django/middleware.py
@@ -1,0 +1,55 @@
+from django.conf import settings
+
+
+class ASAPForwardedMiddleware(object):
+    """Enable client auth for ASAP-enabled services that are forwarding
+    non-ASAP client requests.
+
+    This must come before any authentication middleware.
+    """
+
+    def __init__(self, get_response=None):
+        self.get_response = get_response
+
+        # Rely on this header to tell us if a request has been forwarded
+        # from an ASAP-enabled service; will overwrite X-Forwarded-For
+        self.xfwd = getattr(settings, 'ASAP_PROXIED_FORWARDED_FOR_HEADER',
+                            'HTTP_X_ASAP_FORWARDED_FOR')
+
+        # This header won't always be set, i.e. some users will be anonymous
+        self.xauth = getattr(settings, 'ASAP_PROXIED_AUTHORIZATION_HEADER',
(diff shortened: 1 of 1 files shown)


erikvanzijst:
Is this really what we want? If a request with existing Basic Auth credentials is proxied on to an ASAP-enabled service, shouldn't the original credentials be moved to a different transport header, instead of ASAP?

The contract between ASAP-secured peers is that an ASAP token is supplied through the Authorization header. I'm not sure we need an escaping mechanism for that.

Crazy hypothetical; what if you had another nested of level ASAP-secured proxying in the stack; would the inner pair need to use X-ASAP2 to not overwrite the X-ASAP header from the other proxy layer?

grantisu:
It looks like I forgot to explain what's expected from the ASAP-aware proxy:

When a proxy wants to allow other services to authenticate non-ASAP clients, it needs to put the client's provided Authorization header into the X-ASAP-Authorization header of the proxied request, with the regular ASAP token in the the regular place. Further proxies can do whatever they normally do with the request/ASAP token/etc, so long as the extra X-ASAP-* headers are preserved.

Anonymous clients won't have Authorization headers, and thus the proxy won't create an X-ASAP-Authorization header. In addition, some auth requires the client IP (e.g. IP whitelisting), so we need a way to include that. Since various things change X-Forwarded-For (and related headers) in various ways, the proxy should put the IP it wants to use for the client into the X-ASAP-Forwarded-For header, which shouldn't get mangled.

Since X-ASAP-Authorization won't always be present, we require the X-ASAP-Forwarded-For header to toggle the extra auth behavior.

I think that should answer all the questions you raised; should I update the PR to put that info in a comment/docstring/commit message?

Also: the non-ASAP auth happens _before_ ASAP auth, so it might be difficult for the proxy to distinguish which type of auth caused a failure. Maybe the middleware should generate unauthorized/forbidden responses that mention ASAP...?

erikvanzijst:
Fair enough. LGTM (does GitHub still not have a way to approve a pull request, or am I just not looking in the right place?)

dbaxa:
Please add tests.

PR #59, opened 2017-10-20
Cleanup requires_asap()

grantisu:
The biggest change here is that a server will see exceptions from failed key lookups.

- Allow server to handle server errors
- Raise more consistent exceptions
- Perform some minor code cleanup

### DIFF
--- a/atlassian_jwt_auth/contrib/aiohttp/key.py
+++ b/atlassian_jwt_auth/contrib/aiohttp/key.py
@@ -1,6 +1,7 @@
 import asyncio
 
 import aiohttp
+import requests
 
 from atlassian_jwt_auth.key import (
     PEM_FILE_TYPE,
@@ -21,8 +22,18 @@ class HTTPSPublicKeyRetriever(_HTTPSPublicKeyRetriever):
         return aiohttp.ClientSession(loop=self.loop)
 
     async def _retrieve(self, url, requests_kwargs):
-        resp = await self._session.get(url, headers={'accept': PEM_FILE_TYPE},
-                                       **requests_kwargs)
-        resp.raise_for_status()
-        self._check_content_type(url, resp.headers['content-type'])
-        return await resp.text()
+        try:
+            resp = await self._session.get(url, headers={'accept':
+                                                         PEM_FILE_TYPE},
+                                           **requests_kwargs)
+        except aiohttp.ClientError as e:
+            raise requests.RequestException(e)
+
+        try:
+            resp.raise_for_status()
+            self._check_content_type(url, resp.headers['content-type'])
+            return await resp.text()
+        except aiohttp.http.HttpProcessingError as e:
+            wrapped_exception = requests.HTTPError(str(e))
+            wrapped_exception.original_exception = e
+            raise wrapped_exception
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -2,7 +2,6 @@ from functools import wraps
 
 from django.conf import settings
 from django.http.response import HttpResponse
-from django.utils import six
 
 import atlassian_jwt_auth
 from .utils import parse_jwt, verify_issuers
@@ -19,17 +18,9 @@ def requires_asap(issuers=None):
         def requires_asap_wrapper(request, *args, **kwargs):
             verifier = _get_verifier()
             auth_header = request.META.get('HTTP_AUTHORIZATION', b'')
-            # Per PEP-3333, headers must be in ISO-8859-1 or use an RFC-2047
-            # MIME encoding. We don't really care about MIME encoded
-            # headers, but some libraries allow sending bytes (Django tests)
-            # and some (requests) always send str so we need to convert if
-            # that is the case to properly support Python 3.
-            if isinstance(auth_header, six.string_types):
-                auth_header = auth_header.encode(encoding='iso-8859-1')
-            auth = auth_header.split(b' ')
             err_response = _requires_asap(
                 verifier=verifier,
-                auth=auth,
+                auth=auth_header,
                 parse_jwt_func=parse_jwt,
                 response_class=HttpResponse,
                 asap_claim_holder=request,
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -22,7 +22,7 @@ def requires_asap(issuers=None):
                 verifier=verifier,
                 auth=auth_header,
                 parse_jwt_func=parse_jwt,
-                response_class=HttpResponse,
+                build_response=_build_response,
                 asap_claim_holder=request,
                 verify_issuers_func=verify_issuers,
                 issuers=issuers,
@@ -43,3 +43,14 @@ def _get_verifier():
         base_url=getattr(settings, 'ASAP_PUBLICKEY_REPOSITORY')
     )
     return atlassian_jwt_auth.JWTAuthVerifier(retriever)
+
+
+def _build_response(message, status, headers=None):
+        if headers is None:
+            headers = {}
+
+        response = HttpResponse(message, status=status)
+        for header, value in headers.items():
+            response[header] = value
+
+        return response
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -16,10 +16,9 @@ def requires_asap(issuers=None):
     def requires_asap_decorator(func):
         @wraps(func)
         def requires_asap_wrapper(request, *args, **kwargs):
-            verifier = _get_verifier()
             auth_header = request.META.get('HTTP_AUTHORIZATION', b'')
             err_response = _requires_asap(
-                verifier=verifier,
+                get_verifier=_get_verifier,
                 auth=auth_header,
                 parse_jwt_func=parse_jwt,
                 build_response=_build_response,
--- a/atlassian_jwt_auth/contrib/server/helpers.py
+++ b/atlassian_jwt_auth/contrib/server/helpers.py
@@ -1,5 +1,3 @@
-import logging
-
 from jwt.compat import text_type
 import jwt.exceptions
 
@@ -22,7 +20,7 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response,
     except ValueError:
         scheme = b''
 
-    message, exception = None, None
+    message = None
     if scheme.lower() != b'bearer':
         return build_response('Unauthorized', status=401, headers={
                               'WWW-Authenticate': 'Bearer'})
@@ -33,15 +31,10 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response,
         asap_claim_holder.asap_claims = asap_claims
     except jwt.exceptions.InvalidIssuerError as e:
         message = 'Unauthorized: Invalid token issuer'
-        exception = e
     except jwt.exceptions.InvalidTokenError as e:
         # Something went wrong with decoding the JWT
         message = 'Unauthorized: Invalid token'
-        exception = e
     if message is not None:
-        logger = logging.getLogger(__name__)
-        logger.error(message,
-                     extra={'original_message': str(exception)})
         return build_response(message, status=401, headers={
                               'WWW-Authenticate': 'Bearer'})
     return None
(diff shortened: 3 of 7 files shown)


atlassian_jwt_auth/key.py
```diff
@@ -93,7 +93,10 @@ def retrieve(self, key_identifier, **requests_kwargs):
             key_identifier = KeyIdentifier(key_identifier)
 
         url = self.base_url + key_identifier.key_id
-        return self._retrieve(url, requests_kwargs)
```
dbaxa: I prefer not to wrap exceptions like this unless really needed.

atlassian_jwt_auth/contrib/aiohttp/key.py
```diff
@@ -21,8 +22,16 @@ def _get_session(self):
...
+                                           **requests_kwargs)
+        except aiohttp.ClientError as e:
+            raise requests.RequestException(e)
+
+        try:
+            resp.raise_for_status()
+            self._check_content_type(url, resp.headers['content-type'])
+            return await resp.text()
+        except aiohttp.http.HttpProcessingError as e:
+            raise requests.HTTPError(e)
```
dbaxa: Why wrap this exception?

atlassian_jwt_auth/key.py
grantisu: That would mean that any code that cares about handling failures to retrieve public keys must know all the details about how those keys are received, which is a leaky abstraction.

I'm fine with a more elaborate wrapping scheme than what's currently here, e.g. adding an `original_exception` attribute to `PublicKeyRetrieverException`.

atlassian_jwt_auth/contrib/aiohttp/key.py
grantisu: None of the code that catches key retrieval exceptions or request exceptions handles aiohttp exceptions. Normalizing to one type of exception fixes those bugs and makes for easier code reuse.

I'm not set on `requests.HTTPError` here: we could use something else or create a new exception class.

atlassian_jwt_auth/contrib/aiohttp/key.py
```diff
@@ -21,8 +22,16 @@ def _get_session(self):
         return aiohttp.ClientSession(loop=self.loop)
 
     async def _retrieve(self, url, requests_kwargs):
```
erikvanzijst: Unrelated, but is this library meant to support Python 2?

Edit: never mind, I now see how this is used only on p3k.

atlassian_jwt_auth/contrib/aiohttp/key.py
erikvanzijst: Good catch!

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -30,43 +28,29 @@ def requires_asap_wrapper(request, *args, **kwargs):
...
-                message = 'Unauthorized: Invalid token'
-                exception = e
-            if message is not None:
-                logger = logging.getLogger(__name__)
-                logger.error(message,
-                             extra={'original_message': str(exception)})
-
+                # Invalid token
+                message = 'Unauthorized: %s' % e
                 return HttpResponse(message, status=401)
```
erikvanzijst: Shouldn't we be including a WWW-authenticate response header as per [RFC-6750](https://tools.ietf.org/html/rfc6750#section-3), considering we are using Bearer tokens?

atlassian_jwt_auth/key.py
erikvanzijst: I agree with Grant. This code here should not be coupled to the fact that there is a aiohttp retriever.
Currently the retriever interface exposes `requests` and that's what clients should anticipate. While we *could* abstract the `RequestException` with something implementation-agnostic, I think it wouldn't add too much value.

atlassian_jwt_auth/key.py
```diff
@@ -29,11 +29,11 @@ class KeyIdentifier(object):
     """ This class represents a key identifier """
 
     def __init__(self, identifier):
-        self.__key_id = validate_key_identifier(identifier)
+        self._key_id = validate_key_identifier(identifier)
 
     @property
     def key_id(self):
-        return self.__key_id
+        return self._key_id
```
dbaxa: This change seems unrelated to this pull request.

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -88,16 +88,16 @@ def test_request_with_invalid_audience_is_rejected(self):
             response = self.client.get(reverse('expected'),
                                        HTTP_AUTHORIZATION=b'Bearer ' + token)
 
-        self.assertContains(response, 'Unauthorized: Invalid token',
-                            status_code=401)
+        self.assertContains(response, 'Forbidden',
```
dbaxa: I prefer the previous response message because it provides details on why the request failed. Also, switching from `401` to `403` may seem useful but there seems to be a split about what to do in the case of an authorisation failure amongst asap libraries. For example, the go-asap library uses 401. Unless we have a good reason to change this I am inclined to say stick with `401`.

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -88,16 +88,16 @@ def test_request_with_invalid_audience_is_rejected(self):
...
+                            status_code=403)
 
     def test_request_with_invalid_token_is_rejected(self):
         with override_settings(**self.test_settings):
             response = self.client.get(
                 reverse('expected'),
                 HTTP_AUTHORIZATION=b'Bearer notavalidtoken')
 
-        self.assertContains(response, 'Unauthorized: Invalid token',
```
dbaxa: Why change this?

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -119,7 +119,8 @@ def test_request_with_invalid_issuer_is_rejected(self):
             response = self.client.get(reverse('expected'),
                                        HTTP_AUTHORIZATION=b'Bearer ' + token)
 
-        self.assertContains(response, 'Unauthorized: Invalid token issuer',
```
dbaxa: This doesn't really need to change but I don't mind it.

atlassian_jwt_auth/contrib/aiohttp/key.py
dbaxa: I am not going to agree on this change because you will make it harder for users of the `aiohttp` code to obtain or catch a specific exception that might occur. 

Importantly, 
* `aiohttp.ClientError` extends `Exception` and does not have a `request` nor a `response` attribute.
* `aiohttp.HttpProcessingError` extends `Exception` and has different attributes than `requests.HTTPError`.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -30,43 +28,29 @@ def requires_asap_wrapper(request, *args, **kwargs):
...
-                # Error parsing or getting public key
-                message = 'Unauthorized: Unable to retrieve public key'
-                exception = e
-            except InvalidIssuerError as e:
-                message = 'Unauthorized: Invalid token issuer'
-                exception = e
+            except InvalidAudienceError as e:
+                # Audience mismatch
+                message = 'Forbidden: %s' % e
+                return HttpResponse(message, status=403)
```
dbaxa: I prefer the previous response message because it provides details on why the request failed. Also, switching from `401` to `403` may seem useful but there seems to be a split about what to do in the case of an authorisation failure amongst asap libraries. For example, the go-asap library uses 401. Unless we have a good reason to change this I am inclined to say stick with `401`.

atlassian_jwt_auth/key.py
grantisu: Sorta. This PR started as some minor cleanup, then spiraled out a bit.

I can pull the minor parts out if they're bugging anybody.

atlassian_jwt_auth/contrib/tests/django/test_django.py
grantisu: After re-reading [the spec](https://s2sauth.bitbucket.io/spec/), I think you're right: an invalid audience should result in a `401`, since it kinda makes sense that one server shouldn't recognize credentials that are only valid on another server.

But then we should probably make invalid issuers a `403`. A perfectly valid token from a non-whitelisted issuer/user sounds like an authorization error. Also, in the absence of a subject the issuer is the subject, and rejecting a (properly authenticated) request based on the subject feels like an authorization error.

I'll back out the status code changes in this PR.

atlassian_jwt_auth/contrib/aiohttp/key.py
dbaxa: > If different callers need to have different exceptions, then they probably need to have a custom interface as well.

Right so just to clarify I wasn't thinking that a `django` or` flask` application would use the <django/flask> contrib code with an `asyncio` `HTTPSPublicKeyRetriever`. Perhaps you are thinking of this?

grantisu: I was thinking that `HTTPSPublicKeyRetriever` shouldn't have to care what's calling it, so long as both sides follow the same interface.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -1,45 +1,40 @@
...
-        message = 'Unauthorized: Invalid token issuer'
-        exception = e
+        message = 'Forbidden: Invalid token issuer'
+        return build_response(message, status=403)
     except jwt.exceptions.InvalidTokenError as e:
         # Something went wrong with decoding the JWT
         message = 'Unauthorized: Invalid token'
-        exception = e
-    if message is not None:
-        logger = logging.getLogger(__name__)
```
dbaxa: Why are we removing the logging? (I can moving logging to another location but it is imho useful and this change is just removing it).

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -1,45 +1,40 @@
...
+        return build_response('Unauthorized', status=401, headers={
+                              'WWW-Authenticate': 'Bearer'})
+    verifier = get_verifier()
     try:
-        asap_claims = parse_jwt_func(verifier, auth[1])
+        asap_claims = parse_jwt_func(verifier, auth)
         if verify_issuers_func is not None:
             verify_issuers_func(asap_claims, issuers)
         asap_claim_holder.asap_claims = asap_claims
-    except requests.exceptions.HTTPError as e:
```
dbaxa: Actually this can occur if the key cannot be found and that's not the server's fault at all.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -1,45 +1,40 @@
...
-        # Couldn't find key in key server
-        message = 'Unauthorized: Invalid key'
-        exception = e
-    except (requests.exceptions.ConnectionError,
-            PublicKeyRetrieverException) as e:
-        message = 'Unauthorized: Backend server connection error'
```
dbaxa: It may make sense to make this a non-4xx response.

dbaxa:
I would like to cherry-pick 30dd4a3038dfb657c0d996d1762d2679e849c898 , d434113546ca27a45a34e583404a4f7638575068 6c9e2ed1763e221a9263a8599a942203d9bbfcc4 into a new pull request (with some minor changes).

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: Logging client errors tends to flood the logs with useless entries, especially authorization errors.

For example, a client that only sends authorization headers in response to a 401 will generate an extra "error" log for each request. These should already show up in the request logs, so dropping the logging here shouldn't drop any interesting information.

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: Good point.

This should probably be handled at the point where the request is made: any 4xx response from the key server should raise an `InvalidTokenError` (eventually resulting in a 401), while any other error (can't reach remote server, 5xx response, etc.) should be a `PublicKeyRetrieverException`.

Maybe `InvalidIssuerError` could represent a 4xx response from the key server, and something like `ForbiddenIssuerException` could be a valid issuer that isn't in the whitelist (with a 403 response)?

atlassian_jwt_auth/contrib/server/helpers.py
dbaxa: @grantisu I'd prefer to change the log level (to say debug or info) then drop this logging.

grantisu:
Remaining changes from this PR have been updated/moved to #65

PR #60, opened 2017-10-20
Add validate_asap()

grantisu:
The requires_asap() decorator runs performs all the auth itself, which in Django means that the request must successfully pass through all the middleware before hitting the view. This is good for per-view ASAP validation, but bad for working with other middleware.

The validate_asap() decorator assumes some middleware will check that the ASAP token is valid for the application, and put any resulting claims in request.asap_claims to allow per-endpoint authorization.


grantisu:
Clsoed in favor of #66

PR #61, opened 2017-10-20
Add ASAPMiddleware

grantisu:
Django usually performs authentication in middleware, so that's where we should put ASAP-based authentication.

This is also an opportunity to fix an issue with ASAPForwardedMiddleware, which will unintuitively run other middleware-based authentication before any ASAP authentication.


grantisu:
Closed in favor of #66

PR #63, opened 2017-10-25
Sem-Ver: bugfix Clean up the django and flask requires_asap decorators by sharing their code.

dbaxa:


dbaxa:
cc @grantisu .

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -31,41 +27,18 @@ def requires_asap_wrapper(request, *args, **kwargs):
...
-                # Something went wrong with decoding the JWT
-                message = 'Unauthorized: Invalid token'
-                exception = e
-            if message is not None:
-                logger = logging.getLogger(__name__)
-                logger.error(message,
-                             extra={'original_message': str(exception)})
-
-                return HttpResponse(message, status=401)
+            return result
```
grantisu: Nit: this seems a bit awkward; maybe `s/result/err_response/`?

Alternatively, since we're essentially parameterizing the decorator logic with a helper function, we could change the abstracted function to be a decorator that calls the inner function itself and then have something along the lines of `requires_asap = partial(_requires_asap, **{all the details})` here.

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -17,26 +16,17 @@ def requires_asap(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         verifier = _get_verifier()
```
grantisu: Nit: could `_get_verifier` be passed to `_requires_asap` too?

atlassian_jwt_auth/contrib/server/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+import jwt.exceptions
+import requests.exceptions
+
+from atlassian_jwt_auth.exceptions import (
+    PublicKeyRetrieverException,
+)
+
+
+def _requires_asap(verifier, auth, parse_jwt_func, response_class,
```
grantisu: Nit: `__init__.py` probably shouldn't have code that isn't related to importing.

Maybe just have `server.py` or `server/helpers.py` instead?

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -31,41 +27,18 @@ def requires_asap_wrapper(request, *args, **kwargs):
             if isinstance(auth_header, six.string_types):
                 auth_header = auth_header.encode(encoding='iso-8859-1')
             auth = auth_header.split(b' ')
```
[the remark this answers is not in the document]
dbaxa: I did that originally but then it makes `six` a dependency.

atlassian_jwt_auth/contrib/django/decorators.py
dbaxa: Yeah probably `err_response`.

atlassian_jwt_auth/contrib/flask_app/decorators.py
dbaxa: It could but I didn't see a need to do so.

atlassian_jwt_auth/contrib/server/__init__.py
dbaxa: Sure.

atlassian_jwt_auth/contrib/django/decorators.py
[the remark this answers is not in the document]
grantisu: Could use `jwt.compat.text_type`?

atlassian_jwt_auth/contrib/flask_app/decorators.py
grantisu: Typo on my part: `s/too/instead of verifier/`

Since both decorators already abstract getting the verifier into a function call, we may as well just pass the function to `_requires_asap` directly.

dbaxa: I see no need. Why shouldn't callers be able to provide the verifier?

PR #64, opened 2017-10-26
Fixes for requires_asap()

grantisu:
* Check authorization scheme
* Set `WWW-Authenticate` header


grantisu:
@dbaxa

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -22,7 +22,7 @@ def requires_asap_wrapper(request, *args, **kwargs):
                 verifier=verifier,
                 auth=auth_header,
```
dbaxa: Actually I made this commit but I had left my author information set to grant. Sorry!

PR #65, opened 2017-11-07
Cleanup responses from requires_asap()

grantisu:


atlassian_jwt_auth/contrib/aiohttp/key.py
```diff
@@ -21,8 +22,13 @@ def _get_session(self):
...
+        try:
+            resp = await self._session.get(url, headers={'accept':
+                                                         PEM_FILE_TYPE},
+                                           **requests_kwargs)
+            resp.raise_for_status()
+            self._check_content_type(url, resp.headers['content-type'])
+            return await resp.text()
+        except aiohttp.ClientError as e:
+            status_code = getattr(e, 'code', None)
+            raise PublicKeyRetrieverException(e, status_code=status_code)
```
dbaxa: This is certainly better than the previously proposed change.

atlassian_jwt_auth/exceptions.py
```diff
@@ -1,16 +1,34 @@
...
+        if args:
+            orig = args[0]
+            if isinstance(orig, Exception):
+                args[0] = str(orig)
+                self.original_exception = getattr(orig, 'original_exception',
+                                                  orig)
+        super(_WrappedException, self).__init__(*args, **kwargs)
+
+
+class _WithStatus(object):
```
dbaxa: Some docstring would be nice here.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -32,25 +31,31 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response_func,
         if verify_issuers_func is not None:
             verify_issuers_func(asap_claims, issuers)
         asap_claim_holder.asap_claims = asap_claims
-    except requests.exceptions.HTTPError as e:
+    except PublicKeyRetrieverException as e:
+        if e.status_code != 404:
```
dbaxa: @grantisu actually this is tricky because amazon s3 can reply with `403` when a file is missing.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -32,25 +31,31 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response_func,
...
         message = 'Unauthorized: Invalid token'
         exception = e
     if message is not None:
         logger = logging.getLogger(__name__)
-        logger.error(message,
+        logger.debug(message,
                      extra={'original_message': str(exception)})
-        return build_response_func(message, status=401, headers={
-                              'WWW-Authenticate': 'Bearer'})
+        if message.startswith('Unauthorized:'):
```
dbaxa: Maybe default the response headers to be `Unauthorized` and then check for the `Forbidden` case (or the other way around if you want).

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: Fair enough.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -32,25 +31,34 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response_func,
         if verify_issuers_func is not None:
             verify_issuers_func(asap_claims, issuers)
         asap_claim_holder.asap_claims = asap_claims
-    except requests.exceptions.HTTPError as e:
+    except PublicKeyRetrieverException as e:
+        if e.status_code not in (403, 404):
```
dbaxa: This can be made into a set.

atlassian_jwt_auth/exceptions.py
```diff
@@ -1,16 +1,46 @@
...
     """
 
 
-class PublicKeyRetrieverException(ASAPAuthenticationException):
+class PublicKeyRetrieverException(_WithStatus, ASAPAuthenticationException):
     """Raise when there are issues retrieving the public key"""
 
 
-class PrivateKeyRetrieverException(ASAPAuthenticationException):
+class PrivateKeyRetrieverException(_WithStatus, ASAPAuthenticationException):
```
dbaxa: Why does `PrivateKeyRetrieverException` need `_WithStatus_` ?

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: That would allow an unexpected message to sneak in without throwing an error.

Currently, a bad message will get logged (if debugging is turned on) and then the server will 500 since `kwargs` isn't set. Using a default `kwargs` would allow typos and other unrecognized messages through to the client with no errors.

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: For small collections, there's no meaningful difference between tuples and sets:
```
In [86]: %timeit 403 not in (403, 404)
10000000 loops, best of 3: 20.7 ns per loop

In [87]: %timeit 403 not in {403, 404}
10000000 loops, best of 3: 21.2 ns per loop

In [88]: %timeit 404 not in (403, 404)
10000000 loops, best of 3: 27.8 ns per loop

In [89]: %timeit 404 not in {403, 404}
10000000 loops, best of 3: 28.5 ns per loop

In [90]: %timeit 503 not in (403, 404)
10000000 loops, best of 3: 34.2 ns per loop

In [91]: %timeit 503 not in {403, 404}
10000000 loops, best of 3: 34 ns per loop
```

atlassian_jwt_auth/exceptions.py
grantisu: I guess I was thinking of the "Retriever" part, where a key could conceivably be loaded via HTTPS.

There's obviously no implementation of this for private keys yet, but in terms of API design I think the symmetry with `PublicKeyRetrieverException` makes sense.

I'm fine removing it if it's a blocker, though. ¯\\\_(ツ)\_/¯

atlassian_jwt_auth/contrib/server/helpers.py
dbaxa: Sure. I don't mind.

atlassian_jwt_auth/exceptions.py
dbaxa: I am fine with adding it ... but it doesn't make sense today.

atlassian_jwt_auth/contrib/server/helpers.py
```diff
@@ -32,25 +31,34 @@ def _requires_asap(verifier, auth, parse_jwt_func, build_response_func,
...
         # Couldn't find key in key server
         message = 'Unauthorized: Invalid key'
         exception = e
-    except (requests.exceptions.ConnectionError,
-            PublicKeyRetrieverException) as e:
-        message = 'Unauthorized: Backend server connection error'
-        exception = e
     except jwt.exceptions.InvalidIssuerError as e:
-        message = 'Unauthorized: Invalid token issuer'
+        message = 'Forbidden: Invalid token issuer'
```
dbaxa: My reading of https://s2sauth.bitbucket.io/spec/ suggests that in all cases where we reject the access token we should be responding with a `401`. So shouldn't we be returning a `401` here?

atlassian_jwt_auth/contrib/server/helpers.py
dbaxa: Fair enough.

atlassian_jwt_auth/contrib/server/helpers.py
grantisu: > the access token is accepted but the issuer is not accepted?

Yup: the request has been authenticated, but it doesn't have the right permissions.

The ASAP spec lists RFC 7235 as a reference, but never actually references it. Section 2.1 of RFC 7235 states:

> A server that receives valid credentials that are not adequate to
> gain access ought to respond with the 403 (Forbidden) status code
> (Section 6.5.3 of [RFC7231]).

Section 6.5.3 of RFC 7231 covers 403 responses, about which it says:

> If authentication credentials were provided in the request, the server considers them
> insufficient to grant access.

dbaxa: Thank you for clarifying :-) .

PR #66, opened 2017-11-10
Add better Django ASAP middleware

grantisu:
Django usually performs authentication in middleware, so that's where we should put ASAP-based authentication. But sometimes per-view authorization needs to occur, so we still need a view decorator to handle this.

`ASAPMiddleware` handles ASAP claim validation, including authentication and global authorization.

`validate_asap` handles per-view authorization, and can optionally reject unauthenticated requests for applications that do not want ASAP enabled globally.

This is also an opportunity to fix an issue with `ASAPForwardedMiddleware`, which will unintuitively run other middleware-based authentication before any ASAP authentication.

### DIFF
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -1,10 +1,55 @@
 from functools import wraps
 
 from django.conf import settings
+from django.http.response import HttpResponse
 
 import atlassian_jwt_auth
 from .utils import parse_jwt, verify_issuers, _build_response
 from ..server.helpers import _requires_asap
+
+
+def validate_asap(issuers=None, subjects=None, required=True):
+    """Decorator to allow endpoint-specific ASAP authorization, assuming ASAP
+    authentication has already occurred.
+
+    :param list issuers: A list of issuers that are allowed to use the
+        endpoint.
+    :param subject: A list of subjects or a function to determine allowed
+        subjects for the endpoint.
+    :param boolean required: Whether or not to require ASAP on this endpoint.
+        Note that requirements will be still be verified if claims are present.
+    """
+    def validate_asap_decorator(func):
+        @wraps(func)
+        def validate_asap_wrapper(request, *args, **kwargs):
+            asap_claims = getattr(request, 'asap_claims', {})
+            if required and not asap_claims:
+                message = 'Unauthorized: Invalid or missing token'
+                response = HttpResponse(message, status=401)
+                response['WWW-Authenticate'] = 'Bearer'
+                return response
+
+            iss = asap_claims.get('iss')
+            if issuers and iss not in issuers:
+                message = 'Forbidden: Invalid token issuer'
+                return HttpResponse(message, status=403)
+
+            sub = asap_claims.get('sub')
+            if callable(subjects):
+                sub_allowed = subjects(sub)
+            elif hasattr(subjects, '__contains__'):
+                sub_allowed = sub in subjects
+            else:
+                sub_allowed = True
+
+            if not sub_allowed:
+                message = 'Forbidden: Invalid token subject'
+                return HttpResponse(message, status=403)
+
+            return func(request, *args, **kwargs)
+
+        return validate_asap_wrapper
+    return validate_asap_decorator
 
 
 def requires_asap(issuers=None):
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -22,29 +22,30 @@ def validate_asap(issuers=None, subjects=None, required=True):
     def validate_asap_decorator(func):
         @wraps(func)
         def validate_asap_wrapper(request, *args, **kwargs):
-            asap_claims = getattr(request, 'asap_claims', {})
+            asap_claims = getattr(request, 'asap_claims', None)
             if required and not asap_claims:
                 message = 'Unauthorized: Invalid or missing token'
                 response = HttpResponse(message, status=401)
                 response['WWW-Authenticate'] = 'Bearer'
                 return response
 
-            iss = asap_claims.get('iss')
-            if issuers and iss not in issuers:
-                message = 'Forbidden: Invalid token issuer'
-                return HttpResponse(message, status=403)
+            if asap_claims:
+                iss = asap_claims['iss']
+                if issuers and iss not in issuers:
+                    message = 'Forbidden: Invalid token issuer'
+                    return HttpResponse(message, status=403)
 
-            sub = asap_claims.get('sub')
-            if callable(subjects):
-                sub_allowed = subjects(sub)
-            elif hasattr(subjects, '__contains__'):
-                sub_allowed = sub in subjects
-            else:
-                sub_allowed = True
+                sub = asap_claims.get('sub')
+                if callable(subjects):
+                    sub_allowed = subjects(sub)
+                elif hasattr(subjects, '__contains__'):
+                    sub_allowed = sub in subjects
+                else:
+                    sub_allowed = True
 
-            if not sub_allowed:
-                message = 'Forbidden: Invalid token subject'
-                return HttpResponse(message, status=403)
+                if not sub_allowed:
+                    message = 'Forbidden: Invalid token subject'
+                    return HttpResponse(message, status=403)
 
             return func(request, *args, **kwargs)
 
--- a/atlassian_jwt_auth/contrib/django/decorators.py
+++ b/atlassian_jwt_auth/contrib/django/decorators.py
@@ -14,8 +14,8 @@ def validate_asap(issuers=None, subjects=None, required=True):
 
     :param list issuers: A list of issuers that are allowed to use the
         endpoint.
-    :param subject: A list of subjects or a function to determine allowed
-        subjects for the endpoint.
+    :param list subjects: A list of subjects that are allowed to use the
+        endpoint.
     :param boolean required: Whether or not to require ASAP on this endpoint.
         Note that requirements will be still be verified if claims are present.
     """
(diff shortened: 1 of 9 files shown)


atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -1,14 +1,25 @@
...
+from .utils import parse_jwt, verify_issuers, _build_response
+
 
 class ASAPForwardedMiddleware(object):
     """Enable client auth for ASAP-enabled services that are forwarding
     non-ASAP client requests.
 
     This must come before any authentication middleware.
+
+    DEPRECATED: use ASAPMiddleware instead.
```
dbaxa: Why are we deprecating this ?

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -4,10 +4,54 @@
...
+                message = 'Forbidden: Invalid token issuer'
+                return HttpResponse(message, status=403)
+
+            sub = asap_claims.get('sub')
+            if callable(subjects):
+                sub_allowed = subjects(sub)
+            elif hasattr(subjects, '__contains__'):
+                sub_allowed = sub in subjects
+            else:
+                sub_allowed = True
```
dbaxa: So the default behaviour is to allow the subject.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -4,10 +4,54 @@
...
 from ..server.helpers import _requires_asap
 
 
+def validate_asap(issuers=None, subjects=None, required=True):
+    """Decorator to allow endpoint-specific ASAP authorization, assuming ASAP
+    authentication has already occurred.
+
+    :param list issuers: A list of issuers that are allowed to use the
+        endpoint.
+    :param subject: A list of subjects or a function to determine allowed
```
dbaxa: Why not just accept a list of subjects?

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -4,10 +4,54 @@
...
-from .utils import parse_jwt, verify_issuers
+from .utils import parse_jwt, verify_issuers, _build_response
 from ..server.helpers import _requires_asap
 
 
+def validate_asap(issuers=None, subjects=None, required=True):
+    """Decorator to allow endpoint-specific ASAP authorization, assuming ASAP
+    authentication has already occurred.
+
+    :param list issuers: A list of issuers that are allowed to use the
```
dbaxa: It feels odd to accept a list of `issuers` but then allow a function to be provided for the `subject`. My thoughts on this is what you extract out the `issuer` and the `subject` checks into the django asap classes to allow users to subclass and implement their own checks if that's desired.

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -4,10 +4,54 @@
...
+    :param subject: A list of subjects or a function to determine allowed
+        subjects for the endpoint.
+    :param boolean required: Whether or not to require ASAP on this endpoint.
+        Note that requirements will be still be verified if claims are present.
+    """
+    def validate_asap_decorator(func):
+        @wraps(func)
+        def validate_asap_wrapper(request, *args, **kwargs):
+            asap_claims = getattr(request, 'asap_claims', {})
+            if required and not asap_claims:
```
dbaxa: When would `required` not be true?

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -4,10 +4,54 @@
...
+                message = 'Unauthorized: Invalid or missing token'
+                response = HttpResponse(message, status=401)
+                response['WWW-Authenticate'] = 'Bearer'
+                return response
+
```
dbaxa: I suggest adding a check here for the case of `not required and not asap_claims`.

atlassian_jwt_auth/contrib/django/middleware.py
grantisu: Because it relies on client auth succeeding before ASAP, which is backwards.

atlassian_jwt_auth/contrib/django/decorators.py
grantisu: Yup: if no restrictions are provided, then we shouldn't restrict access beyond the required fields.

atlassian_jwt_auth/contrib/django/decorators.py
grantisu: Issuers should be a relatively small list of other services, but subjects are whoever the service is acting for. If subjects were explicitly restricted to other services, then a list would make sense, but I can't find that in the spec.

A possible use case would be to have a service make a call on behalf of a user, where subject is set to an email address, a UUID, or something else that requires a function call to validate.

grantisu: I think I misunderstood your comment: pulling checks out into a class that could be overridden sounds fine, but I don't think there's currently a good place to do that.

Adding something like a `ClaimsValidator` class that could be used by the middleware and view decorators sounds fine to me, but out of scope for this PR.

So I can strip out the subjects function stuff for now, assuming a more comprehensive mechanism will be added later?

dbaxa: I don't mind keeping it in. The problem I have is that it is adding complexity where we probably don't need it right now.

grantisu: Yeah, I'll just remove it.

Also: any opinions on the `ClaimsValidator` approach?

dbaxa: Sounds okay to me but it might not be needed - re class methods.

atlassian_jwt_auth/contrib/django/decorators.py
grantisu: I addressed this in fea56c1829e83497fea0b74d620116e89da545c7; are there any other issues preventing this from being merged?

atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -53,3 +66,44 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+
+class ASAPMiddleware(ASAPForwardedMiddleware):
+    """Enable ASAP for Django applications.
+
+    To use proxied credentials, this must come before any authentication
+    middleware.
+    """
+
+    def __init__(self, get_response=None):
+        with warnings.catch_warnings():
```
dbaxa: I disagree with the way that `ASAPForwardedMiddleware` is deprecated but then the warning is captured here. Others who have extended `ASAPForwardedMiddleware` should not be forced to do the same.

I guess what I am saying is that I don't feel we should deprecate `ASAPForwardedMiddleware` at this point in time or if we want to do so we shouldn't be emitting a warning in the __init__(...) of `ASAPForwardedMiddleware`.

grantisu: > Others who have extended ASAPForwardedMiddleware should not be forced to do the same.

My plan was to either eliminate `ASAPForwardedMiddleware` or mark it private in a future PR, and have anybody using/extending it get a deprecation warning so they could plan accordingly.

> I guess what I am saying is that I don't feel we should deprecate ASAPForwardedMiddleware at this point in time [...]

I'd rather deprecate it early than risk having additional people rely on it.

> [I]f we want to do so we shouldn't be emitting a warning in the init(...) [...]

I prefer the warning, but only by a little bit. I'm strongly in favor of *some* form of deprecation; would dropping the warning and leaving the deprecation in the doc string be sufficient? Or is the warning just in the wrong place?

dbaxa: Yep. That's fine my disagreement is mainly with the issued `warning`.

atlassian_jwt_auth/contrib/django/utils.py
```diff
@@ -32,3 +33,14 @@ def verify_issuers(asap_claims, issuers=None):
         logger.error(message, extra={'iss': claim_iss})
 
         raise InvalidIssuerError(message)
+
+
+def _build_response(message, status, headers=None):
```
dbaxa: Was the reason behind moving `_build_response` out to here?

dbaxa:
Please pull in the pep8 fixes I performed in https://github.com/atlassian/asap-authentication-python/tree/django-asap-middleware-from-gmathews .

Also, before merging these changes I require tests are added for the new code.

atlassian_jwt_auth/contrib/django/decorators.py
dbaxa: Yes some pep8 fixes and the addition of tests to check the new code.

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -19,6 +19,124 @@ def create_token(issuer, audience, key_id, private_key):
     return signer.generate_jwt(audience)
 
 
+@modify_settings(MIDDLEWARE={
+    'prepend': 'atlassian_jwt_auth.contrib.django.middleware.ASAPMiddleware',
+})
+class TestAsapMiddleware(RS256KeyTestMixin, SimpleTestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
dbaxa: Can we avoid duplication of code here by sharing the similar aspects of `TestAsapDecorator` and `TestAsapMiddleware` ? (e.g. `setUpClass`, `tearDownClass`, `setUp`)

atlassian_jwt_auth/contrib/tests/django/test_django.py
```diff
@@ -49,6 +50,98 @@ def setUp(self):
             'ASAP_KEY_RETRIEVER_CLASS': self.retriever
         }
 
+
+@modify_settings(MIDDLEWARE={
+    'prepend': 'atlassian_jwt_auth.contrib.django.middleware.ASAPMiddleware',
+})
+class TestAsapMiddleware(DjangoAsapMixin, RS256KeyTestMixin, SimpleTestCase):
```
dbaxa: We are missing tests for checking the `subject` where only specific `subjects` are allowed.

PR #71, opened 2017-12-22
Sem-Ver: bugfix Catch PublicKeyRetrieverException inside _requires_asap regardless of the status code of the exception.

dbaxa:


dbaxa:
cc @grantisu

PR #75, opened 2018-03-22
Sem-Ver: feature Support disabling checking if jwt jti are unique.

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/verifier.py
+++ b/atlassian_jwt_auth/verifier.py
@@ -80,7 +80,6 @@ class JWTAuthVerifier(object):
         """Checks that the given jti has not been already been used."""
         if jti in self._seen_jti:
             raise ValueError("The jti, '%s', has already been used." % jti)
-        else:
-            self._seen_jti[jti] = None
-            while len(self._seen_jti) > 1000:
-                self._seen_jti.popitem(last=False)
+        self._seen_jti[jti] = None
+        while len(self._seen_jti) > 1000:
+            self._seen_jti.popitem(last=False)
--- a/atlassian_jwt_auth/verifier.py
+++ b/atlassian_jwt_auth/verifier.py
@@ -16,6 +16,8 @@ class JWTAuthVerifier(object):
         self._seen_jti = OrderedDict()
         self._subject_should_match_issuer = kwargs.get(
             'subject_should_match_issuer', True)
+        self._check_jti_uniqueness = kwargs.get(
+            'check_jti_uniqueness', True)
 
     def verify_jwt(self, a_jwt, audience, leeway=0, **requests_kwargs):
         """Verify if the token is correct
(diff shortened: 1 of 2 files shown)


mark-adams:
As a general comment on this approach, many Python web applications usually use a worker process-based concurrency model and this mechanism of doing in-memory JTI uniqueness is not really useful in that regard since an app may have dozens of worker processes running at any given time.

At some point, it might be nice to abstract this a little more and allow for the caller to specify a backend callable that can be used to check for JTI uniqueness. Doing so would allow consumers to use Redis or Memcache as a backing store and truly enforce JTI uniqueness across the app.

atlassian_jwt_auth/verifier.py
```diff
@@ -69,12 +71,16 @@ def _decode_jwt(self, a_jwt, key_identifier, jwt_key,
...
+            self._check_jti(_jti)
+        return claims
+
+    def _check_jti(self, jti):
+        """Checks that the given jti has not been already been used."""
+        if jti in self._seen_jti:
+            raise ValueError("The jti, '%s', has already been used." % jti)
         else:
-            self._seen_jti[_jti] = None
+            self._seen_jti[jti] = None
```
mark-adams: I don't really understand `self._seen_jti` at all. This should most certainly be a `set()` instead of an ordered dictionary.

atlassian_jwt_auth/verifier.py
```diff
@@ -16,6 +16,8 @@ def __init__(self, public_key_retriever, **kwargs):
         self._seen_jti = OrderedDict()
         self._subject_should_match_issuer = kwargs.get(
             'subject_should_match_issuer', True)
+        self._check_jti_uniqueness = kwargs.get(
```
mark-adams: I think if we continue going down the `kwargs` route, this begins to get hairy from a consumer point of view because you actually have to dive in here and read the code to discover how to properly call the function. I'd much rather eliminate `**kwargs` here and see something like:

```
def __init__(self, retriever, subject_should_match_issuer=True, check_jti_uniqueness=True):
    ...
```

mark-adams: How about we call this `require_unique_jti`?

mark-adams:
Is somebody asking for this? It seems odd that a consumer would even include a `jti` if it wasn't intending to avoid token reuse...

atlassian_jwt_auth/verifier.py
dbaxa: Sure.

dbaxa:
> Is somebody asking for this? It seems odd that a consumer would even include a jti if it wasn't intending to avoid token reuse...

Well they would need to provide a value for `jti` because of how we access it if it wasn't provided a `KeyError` would be raised.

atlassian_jwt_auth/verifier.py
dbaxa: @mark-adams it was originally a set but @erikvanzijst proposed that we change it to an `OrderedDict` so that we would remove the specific entries (#57).

atlassian_jwt_auth/verifier.py
dbaxa: That's true but I think if we document the keyword arguments it can take it might be okay.

PR #79, opened 2018-04-11
Reuse JWTs within the time window that they are valid

bedwards:
This gives an optional performance boost compared to generating a JWT token with each request. The default behavior is preserved. This is inline with the spec...

> The client MAY reuse the same token for multiple HTTP requests to the resource server within the lifetime of the token except if the resource server explicitly documented that it enforces a token single use policy.

https://s2sauth.bitbucket.io/spec/


dberrueta-atlassian:
I'm not familiar with this library, but please make sure (ideally with a test) that *any* change to any of the claims or headers of the JWT token, including custom claims, cause a cache miss and a new token is generated.

atlassian_jwt_auth/auth.py
```diff
@@ -1,24 +1,47 @@
 from __future__ import absolute_import
+from datetime import datetime, timedelta
 
 import atlassian_jwt_auth
 
+# Regenerate the JWT token once 95% of its valid time window has passed
+_DEFAULT_REUSE_JWT_THRESHOLD = 0.95
```
dbaxa: My personal preference is to avoid using globals as much as possible.

atlassian_jwt_auth/auth.py
```diff
@@ -1,24 +1,47 @@
...
+            self._encoded_jwt, self._claims = self._signer.generate_jwt(
+                self._audience, additional_claims=self._additional_claims,
+                return_claims=True)
+        return b'Bearer ' + self._encoded_jwt
+
+    def _should_generate_jwt(self):
+        if not self._reuse_jwts or self._encoded_jwt is None:
+            return True
+
+        # Rengerate the JWT if it is about to expire
```
dbaxa: Regenerate?

atlassian_jwt_auth/signer.py
```diff
@@ -43,15 +43,15 @@ def _generate_claims(self, audience, **kwargs):
...
-        return jwt.encode(
-            self._generate_claims(audience, **kwargs),
-            key=private_key_pem,
-            algorithm=self.algorithm,
-            headers={'kid': key_identifier.key_id})
+        claims = self._generate_claims(audience, **kwargs)
+        encoded_jwt = jwt.encode(dict(claims), key=private_key_pem,
+                                 algorithm=self.algorithm,
+                                 headers={'kid': key_identifier.key_id})
+        return (encoded_jwt, claims) if return_claims else encoded_jwt
```
dbaxa: What's the performance penalty on decoding without verifying/validating the generated jwt ? (Also, I have a strong preference for making methods always return results in the same form - so I'd prefer to change this to always return a tuple of claims and the generated jwt).

dbaxa:
I am not sure this approach is the one we should be taking regarding reusing valid jwt that have matching claims. While I would prefer to add an addition class (and methods to it) to support user's regenerating tokens & allow them to subclass the base "reuse token" class, that will require more changes. Essentially, my concern is that if user's of this library want to share generated tokens between nodes or store generated tokens in an application specific cache (e.g. a `django` cache) that should also be possible.

mark-adams:
> Essentially, my concern is that if user's of this library want to share generated tokens between nodes or store generated tokens in an application specific cache (e.g. a django cache) that should also be possible

We ran into a similar problem in the PyJWT library with `jti` claims validation. Should we implement this in-memory or use an external cache? Do we need to build pluggable backends? What backends should we build? etc. In the end, we decided to punt completely on `jti` validation and leave it up to the applications to implement their own solution.

atlassian_jwt_auth/signer.py
bedwards: If always returning a tuple won't break code that uses this library then I am all for doing that.

I am not sure about the performance of decoding, but it doesn't sit right with me.

bedwards:
To allow for extension of the token reuse policy (e.g. sharing between nodes, using a framework's cache system), I created a TokenReusePolicy class and pass that into the BaseJWTAuth constructor. The create classmethod retains the simple interface of reuse_jwts=boolen, reuse_jwt_threshold=float.

bedwards:
@dberrueta-atlassian , Python isn't great at guaranteeing this, but basically the design of the BaseJWTAuth and JWTAuthSigner classes is for the claims to only be set in the constructors and never changed through the lifetime of the instance. BaseJWTAuth uses the convention of leading underscore. JWTAuthSigner should probably do the same thing.

PR #80, opened 2018-04-18
Support reusing tokens

dbaxa:


dbaxa:
@bedwards - I'll note that I also thought about another approach where a wrapper class of `JWTAuthSigner` could use a `ReuseTokenMixin` (it could also use composition instead of inheritance re `JWTAuthSigner`).

atlassian_jwt_auth/signer.py
```diff
@@ -54,10 +55,59 @@ def generate_jwt(self, audience, **kwargs):
...
+            issuer, private_key_retriever, **kwargs)
+        self.reuse_threshold = kwargs.get('reuse_jwt_threshold', 0.95)
+
+    def get_cached_token(self):
+        return getattr(self, '_previous_token', None)
+
+    def set_cached_token(self, value):
+        self._previous_token = value
+
+    def can_reuse_token(self, existing_token, claims):
```
dbaxa: We can probably extract this method (but would want to add checks on the key identifier if we do that).

atlassian_jwt_auth/signer.py
```diff
@@ -54,10 +55,59 @@ def generate_jwt(self, audience, **kwargs):
...
+        if existing_token is None:
+            return False
+        existing_claims = jwt.decode(existing_token, verify=False)
```
dbaxa: Technically we don't need to call `jwt.decode` and we can modify the code to set the claims in addition to the token.

atlassian_jwt_auth/contrib/tests/test_requests.py
```diff
@@ -69,6 +70,72 @@ def test_create_jwt_auth_with_additional_claims(self):
...
+
+    def test_do_not_reuse_jwt_if_issuer_changes(self):
+        auth = self.create_jwt_auth('issuer', 'issuer/key',
+                                    self._private_key_pem.decode(), 'audience',
+                                    algorithm=self.algorithm, reuse_jwts=True)
+        auth_header = self._get_auth_header(auth)
+        auth._signer.issuer = 'not-' + auth._signer.issuer
+        self.assertNotEqual(auth_header, self._get_auth_header(auth))
+
+    def test_do_not_reuse_jwt_if_lifetime_changes(self):
```
dbaxa: Do we want to enforce this?

atlassian_jwt_auth/signer.py
```diff
@@ -54,15 +55,73 @@ def generate_jwt(self, audience, **kwargs):
...
+        if calendar.timegm(self._now().utctimetuple()) > about_to_expire:
+            return False
+        if set(claims.keys()) != set(existing_claims.keys()):
+            return False
+        for key, val in claims.items():
+            if key in ['exp', 'iat', 'jti', 'nbf']:
+                continue
+            if existing_claims[key] != val:
+                return False
+        return True
```
bedwards: This looks good to me. I agree it makes sense to keep the claims inside the Signer and not expose them to the Auth instance.

PR #82, opened 2018-05-02
Rework Django and Flask support

mark-adams:
I'm submitting this PR before it's totally finished to get some initial feedback. Remaining work includes:

- [x] Adding documentation for public APIs
- [x] Improving test coverage

Previously, contributors introduced Django and Flask support via
differing implementations into the contrib/ package. The implementations
were similar (sometimes almost copied and pasted from one another) but
also diverged in terms of their functionality and feature sets.

This change's primary goal is to take the contrib/{django,flask} and
refactor them into a shared implementation with framework-specific
backends to abstract away the differences between the two. To
demonstrate this formalization of support for these frameworks, they
have also been moved from contrib into frameworks/{django,flask}.

In addition the following changes were made in this commit:

- Django and Flask decorators no longer require an issuer to be present
in BOTH the global setting and the decorator argument to be valid.
Decorator arguments now override the global setting.
- The proxied middleware seems like a one-off special case and remains
in contrib while a more general ASAP middleware which applies the
decorator logic to all views is included in the supported package.
- The @require_asap decorator has been renamed to @with_asap since you
can technically pass the required=False argument to make ASAP not
enforced.
- The @validate_asap decorator has been renamed to a more descriptive
@restrict_asap to symbolize that it applies additional restrictions onto
an already processed ASAP authenticated request.
- The 'subjects' arg to the @validate_asap decorator has been removed
since it appears to be unused and is inconsistent with the existing
@with_asap decorator.

Sem-Ver: api-break

It should be noted that this is a breaking change in both functionality and API for the Django / Flask  integrations but I think it is worth it to clean up some of the tech debt and simplify the implementation into a common well-tested model.


atlassian_jwt_auth/frameworks/flask/backend.py
```diff
@@ -0,0 +1,41 @@
+from flask import Response, current_app, g, request as current_req
+
+from ..common.backend import Backend
+from ..common.utils import SettingsDict
+
+
+class FlaskBackend(Backend):
+    def get_authorization_header(self, request=None):
+        if request is None:
+            request = current_req
```
dbaxa: What if this is still `None`?

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -0,0 +1,95 @@
...
+        # The amount of leeway to apply when evaluating token expiration
+        # timestamps
+        'ASAP_VALID_LEEWAY': 0,
+
+        # A list of valid token issuers allowed to authenticate
+        # (this can be overridden at the decorator level)
+        'ASAP_VALID_ISSUERS': None,
+
+        # Enforce that the ASAP subject must match the issuer
+        'ASAP_SUBJECT_SHOULD_MATCH_ISSUER': False
```
dbaxa: Shouldn't we default this to `None`?

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -0,0 +1,95 @@
...
+        'ASAP_REQUIRED': True,
+
+        # The valid audience valuw expected when authenticating tokens
+        'ASAP_VALID_AUDIENCE': None,
+
+        # The amount of leeway to apply when evaluating token expiration
+        # timestamps
+        'ASAP_VALID_LEEWAY': 0,
+
+        # A list of valid token issuers allowed to authenticate
```
dbaxa: Ideally this would be a set.

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -0,0 +1,95 @@
...
+        'ASAP_KEY_RETRIEVER_CLASS': HTTPSPublicKeyRetriever,
+
+        # The repository URL where the key retriever can fetch public keys
+        'ASAP_PUBLICKEY_REPOSITORY': None,
+
+        # Whether or not ASAP authentication is required
+        # This is primarily useful when phasing in ASAP authentication
+        'ASAP_REQUIRED': True,
+
+        # The valid audience valuw expected when authenticating tokens
```
dbaxa: Typo here (`valuw`).

atlassian_jwt_auth/frameworks/common/decorators.py
```diff
@@ -0,0 +1,79 @@
...
+def _with_asap(func=None, backend=None, issuers=None, required=None,
+               subject_should_match_issuer=None):
+    if backend is None:
+        raise ValueError(
+            'Invalid value for backend. Use a subclass instead.'
+        )
+
+    def with_asap_decorator(func):
+        @wraps(func)
+        def with_asap_wrapper(*args, **kwargs):
```
dbaxa: Why not request a `request` here as an argument?

atlassian_jwt_auth/frameworks/flask/backend.py
mark-adams: If that happens, one of the fundamental concepts in Flask is broken since current_req is a global value set by Flask that always contains the current request who's lifecycle you are inside of.

To be exact, if this is still `None`, you're not using it inside of a Flask request context and you'll get exceptions.

In reality, as long as you are only using the FlaskBackend with Flask, that simply won't happen.

The only reason this `if` is here at all instead of simply using `current_req` is because I wanted to allow us to pass in an override request during testing.

atlassian_jwt_auth/frameworks/common/backend.py
mark-adams: No, I think that would be more ambiguous. It's clear to the consumer here that if you don't override the config, the subject does not have to match the issuer. If we set it to `None`, it wouldn't be clear... is it on or is it off?

atlassian_jwt_auth/frameworks/common/backend.py
mark-adams: Agreed. I'll change it to say an `iterable` and then create a set inside when constructing the final settings dict.

atlassian_jwt_auth/frameworks/common/backend.py
mark-adams: Yep, I'll fix that. Good find.

atlassian_jwt_auth/frameworks/common/decorators.py
mark-adams: Django passes the `request` to the view as an `arg`; not a `kwarg` and Flask doesn't pass it at all instead preferring the global request object that is automatically set.

atlassian_jwt_auth/frameworks/django/backend.py
```diff
@@ -0,0 +1,51 @@
...
+        response = HttpResponseForbidden(data)
+        for k, v in headers.items():
+            response[k] = v
+
+        return response
+
+    def set_asap_claims_for_request(self, request, claims):
+        request.asap_claims = claims
+
+    @property
```
dbaxa: So settings should not change?

atlassian_jwt_auth/frameworks/django/decorators.py
```diff
@@ -0,0 +1,51 @@
+from ..common.decorators import _with_asap, _restrict_asap
+from .backend import DjangoBackend
+
+
+def with_asap(func=None, issuers=None, required=None,
```
dbaxa: Do we support a `None` value for `func`?

atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -1,21 +1,17 @@
...
-from ..server.helpers import _requires_asap
-from .utils import parse_jwt, verify_issuers, _build_response
-from .decorators import _get_verifier
+from atlassian_jwt_auth.frameworks.django.middleware import (
+    OldStyleASAPMiddleware
+)
 
 
-class ASAPForwardedMiddleware(object):
+class ProxiedAsapMiddleware(OldStyleASAPMiddleware):
```
dbaxa: Why are we changing the name :-) ?

atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -1,21 +1,17 @@
...
     """Enable client auth for ASAP-enabled services that are forwarding
     non-ASAP client requests.
 
-    This must come before any authentication middleware.
-
-    DEPRECATED: use ASAPMiddleware instead.
```
dbaxa: So this is no longer deprecated?

atlassian_jwt_auth/frameworks/common/backend.py
dbaxa: My preference is for to take on the value that we use as the default in `JWTAuthVerifier`.

mark-adams: What are your thoughts on making this the default? It seems silly to, by default, require ASAP tokens to include a duplicative subject key for no other reason than to match the issuer.

atlassian_jwt_auth/frameworks/common/utils.py
```diff
@@ -0,0 +1,94 @@
...
+    def __getattr__(self, name):
+        if name not in self:
+            raise AttributeError
+
+        return self[name]
+
+
+def process_asap_token(request, backend, issuers=None, required=None,
+                       subject_should_match_issuer=None):
+    verifier = backend.get_verifier()
```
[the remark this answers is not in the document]
mark-adams: It's generally better to pass only what you need to later functions. In this case, the later functions (`_check_asap_token()`) doesn't need `backend`, it just needs the token, verifier, and settings.

atlassian_jwt_auth/frameworks/common/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+class SettingsDict(dict):
+    def __getattr__(self, name):
+        if name not in self:
+            raise AttributeError
+
+        return self[name]
+
+
+def process_asap_token(request, backend, issuers=None, required=None,
```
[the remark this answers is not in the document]
mark-adams: Yea, good call. I just pushed up a commit that duplicates and modifeis the session objects in the decorators so now only `request`, `backend`, and `settings` get passed in. Much better!

atlassian_jwt_auth/frameworks/django/backend.py
```diff
@@ -0,0 +1,51 @@
+from django.conf import settings as django_settings
+from django.http import HttpResponse, HttpResponseForbidden
+from ..common.backend import Backend
+from ..common.utils import SettingsDict
+
+
+class DjangoBackend(Backend):
+    def get_authorization_header(self, request=None):
+        if request is None:
```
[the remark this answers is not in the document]
mark-adams: If you're referring to the optional `request=None` argument above, I left that to maintain the exact same API as the base Backend.
Otherwise, that's essentially what I'm doing, right? I'm checking to see if `request is None` and if it is, I'm raising an exception.

atlassian_jwt_auth/frameworks/django/backend.py
mark-adams: Outside of test situations, the settings provided by a backend should be the same no matter how many times you call this.

atlassian_jwt_auth/contrib/django/middleware.py
mark-adams: I think there's two use cases to talk about here:
- Django middleware which performs ASAP auth on all requests 
- Django middleware with weird client proxy behavior

Previously, the two were combined into the same middleware (which had two versions) which made the first and more common case unnecessarily complicated. I split out the simple use-case into an old-style and new-style middleware in `frameworks/django`. I then refactored and combined the two middleware in `contrib/django/middleware.py` into a single `ProxiedAsapMiddleware` that inherits from the simple middleware in `frameworks/django`.

The main idea here is... I flattened the odd contrib custom middleware (as long as we're cleaning up this tech debt anyways) and created a new simple middleware based off the shared common decorator logic in frameworks/django.

So, to answer your original question, the deprecated middleware doesn't really exist anymore but the diff makes it look more confusing than it is.

atlassian_jwt_auth/contrib/django/middleware.py
dbaxa: Should we make use of `django.utils.deprecation.MiddlewareMixin` ?

mark-adams: Good point. I didn't think about this middleware being used by Django 1.11+ consumers. Fixed!

mark-adams:
@dbaxa Can you approve this if there are no remaining blockers?

atlassian_jwt_auth/frameworks/django/tests/test_django.py
```diff
@@ -232,22 +222,6 @@ def test_request_with_invalid_issuer_is_rejected(self):
         self.assertContains(response, 'Forbidden: Invalid token issuer',
                             status_code=403)
 
-    def test_request_non_whitelisted_decorated_issuer_is_rejected(self):
```
dbaxa: Hmm should we restore this test case?

mark-adams:
@dbaxa I went ahead and added Chris (@PossibleSloth) on this PR to try and get a second set of eyes on it as well.

I think we've resolved all of your outstanding feedback. Would it be possible to unblock my project work and get this merged since it seems like all comments are resolved and all tests are passing? Ideally, I'd like to do this tomorrow morning if possible.

Thanks to both of you for the reviews!

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -0,0 +1,102 @@
...
+        # The amount of leeway to apply when evaluating token expiration
+        # timestamps
+        'ASAP_VALID_LEEWAY': 0,
+
+        # An iterable of valid token issuers allowed to authenticate
+        # (this can be overridden at the decorator level)
+        'ASAP_VALID_ISSUERS': None,
+
+        # Enforce that the ASAP subject must match the issuer
+        'ASAP_SUBJECT_SHOULD_MATCH_ISSUER': None
```
PossibleSloth: Since this is a boolean, shouldn't the default be either "True" or "False"?

atlassian_jwt_auth/frameworks/common/exceptions.py
```diff
@@ -0,0 +1,5 @@
+from atlassian_jwt_auth.exceptions import ASAPAuthenticationException
+
+
+class NoTokenProvidedError(ASAPAuthenticationException):
```
dbaxa: Well arguably our existing exceptions inherit from `ValueError` (https://github.com/atlassian/asap-authentication-python/blob/master/atlassian_jwt_auth/exceptions.py)

atlassian_jwt_auth/frameworks/common/backend.py
dbaxa: If we are going to change the default behaviour, let's do it in another pr and not this one :-).

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -44,6 +45,11 @@ def validate_asap_wrapper(request, *args, **kwargs):
 
 
 def requires_asap(issuers=None, subject_should_match_issuer=None, func=None):
+    warnings.warn(
+        "requires_asap in the contrib package is deprecated;"
+        "use atlassian_jwt_auth.frameworks.django.requires_asap instead",
```
dbaxa: `Please use`?

atlassian_jwt_auth/contrib/django/decorators.py
```diff
@@ -44,6 +45,11 @@ def validate_asap_wrapper(request, *args, **kwargs):
 
 
 def requires_asap(issuers=None, subject_should_match_issuer=None, func=None):
+    warnings.warn(
+        "requires_asap in the contrib package is deprecated;"
```
dbaxa: as `deprecated as of version 4.0.0` ?

atlassian_jwt_auth/contrib/flask_app/decorators.py
```diff
@@ -1,7 +1,14 @@
+import warnings
+
 from atlassian_jwt_auth.frameworks.flask.decorators import with_asap
 
 
 def requires_asap(f, issuers=None, subject_should_match_issuer=None):
+    warnings.warn(
```
dbaxa: perhaps` deprecated as of version 4.0.0;` ?

atlassian_jwt_auth/frameworks/common/exceptions.py
mark-adams: Sure. I'm not strongly opposed to it. I only put it here because it is thrown and caught only within this package.

atlassian_jwt_auth/frameworks/wsgi/tests/test_wsgi.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def create_token(issuer, audience, key_id, private_key):
+    signer = atlassian_jwt_auth.create_signer(
+        issuer, key_id, private_key
+    )
+    return signer.generate_jwt(audience)
+
+
+class WsgiTests(utils.RS256KeyTestMixin, unittest.TestCase):
```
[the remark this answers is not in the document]
mark-adams: Perhaps. I'm generally a fan of having crypto backend modules similar to how we have Django, Flask, and WSGI backends here and then having good unit tests for all the backends and pick the most common one to be used in the integration tests.

To be honest, I'd really like to see us revamp all these tests to be PyTest parameterized tests instead but that's as separate PR.

atlassian_jwt_auth/contrib/django/decorators.py
mark-adams: Technically, it's deprecated immediately as of the moment it's committed. :+1: 
If anything, we should be telling people when it will be discontinued completely.

atlassian_jwt_auth/contrib/flask_app/decorators.py
mark-adams: Technically, it's deprecated immediately as of the moment it's committed. :+1: 
If anything, we should be telling people when it will be discontinued completely.

PR #86, opened 2018-05-09
Rework frameworks follow up

dbaxa:


atlassian_jwt_auth/frameworks/common/decorators.py
```diff
@@ -0,0 +1,105 @@
...
+               subject_should_match_issuer=None):
+    if backend is None:
+        raise ValueError(
+            'Invalid value for backend. Use a subclass instead.'
+        )
+
+    def with_asap_decorator(func):
+        @wraps(func)
+        def with_asap_wrapper(*args, **kwargs):
+            settings = _update_settings_from_kwargs(
```
dbaxa: This part is duplicated here and inside `restrict_asap_decorator`

atlassian_jwt_auth/verifier.py
```diff
@@ -69,12 +71,15 @@ def _decode_jwt(self, a_jwt, key_identifier, jwt_key,
...
-            raise ValueError("The jti, '%s', has already been used." % _jti)
-        else:
-            self._seen_jti[_jti] = None
-            while len(self._seen_jti) > 1000:
-                self._seen_jti.popitem(last=False)
+        if self._check_jti_uniqueness:
+            self._check_jti(_jti)
         return claims
+
+    def _check_jti(self, jti):
```
mark-adams: I love this change 👍

dbaxa: ?

PR #87, opened 2018-05-25
Rework frameworks follow up v2

dbaxa:


atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -81,17 +81,21 @@ def get_asap_token(self, request):
...
+    def get_verifier(self, settings=None):
         """Returns a verifier for ASAP JWT tokens"""
-        settings = self.settings
+        if settings is None:
+            settings = self.settings
 
         retriever = settings.ASAP_KEY_RETRIEVER_CLASS(
             base_url=settings.ASAP_PUBLICKEY_REPOSITORY
         )
+        kwargs = {}
```
dbaxa: Hmm we might be missing a test to check when the subject and issuer differ for a token - that we reject that token.

atlassian_jwt_auth/frameworks/common/exceptions.py
```diff
@@ -1,6 +1,2 @@
 class NoTokenProvidedError(Exception):
     pass
```
dbaxa: Should `NoTokenProvidedError` extend `ValueError`?

atlassian_jwt_auth/frameworks/common/backend.py
dbaxa: I have added a test to cover the case I was thinking about.

atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -2,7 +2,8 @@
 from django.utils.deprecation import MiddlewareMixin
 
 from atlassian_jwt_auth.frameworks.django.middleware import (
-    OldStyleASAPMiddleware
+    OldStyleASAPMiddleware,
+    asap_middleware as ASAPMiddleware,
```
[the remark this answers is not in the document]
mark-adams: I'm not a fan of this particular change because Django tends to follow the normal Python convention of `ClassName` and `function_name` when distinguishing between the old-style and new-styles middleware.

I don't mind importing `asap_middleware` but I'd rather not call it `ASAPMiddleware` since I think that will confuse callers.

atlassian_jwt_auth/frameworks/common/backend.py
mark-adams: Sounds great!

atlassian_jwt_auth/frameworks/common/exceptions.py
mark-adams: I think inheriting from `Exception` is fine even though we probably should have a common base exception (in a separate PR) that all of our exceptions inherit from.

dbaxa: Well arguably our existing exceptions inherit from `ValueError` (https://github.com/atlassian/asap-authentication-python/blob/master/atlassian_jwt_auth/exceptions.py)

PR #89, opened 2018-06-26
frameworks: Add support for ASAP_EXCLUDE_PATHS to exlcude paths from ASAP

mark-adams:
While integrating with one of our services, I realized that I didn't provide any mechanism in the middleware to allow an application author to exclude paths from the middleware.

This change adds a new `ASAP_EXCLUDE_PATHS` setting that allows an author to specify a list of regular expressions (as strings) that will be used to determine if a path should be excluded or not.

### DIFF
--- a/atlassian_jwt_auth/frameworks/common/asap.py
+++ b/atlassian_jwt_auth/frameworks/common/asap.py
@@ -14,6 +14,13 @@ def _process_asap_token(request, backend, settings):
     if token is None and not settings.ASAP_REQUIRED and (
             settings.ASAP_REQUIRED is not None):
         return
+
+    if request and settings.ASAP_EXCLUDE_PATHS:
+        path = request.environ.get('PATH_INFO')
+        excluded_paths = settings.ASAP_EXCLUDE_PATHS
+        if any([excluded.match(path) for excluded in excluded_paths]):
+            return
+
     try:
         if token is None:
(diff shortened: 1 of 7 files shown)


dbaxa:
I am not a fan and have never been a fan of excluding specific paths from being checked. I strongly suggest not requiring ASAP (if asap is not to be required globally in an application) and then using the appropriate decorators.

mark-adams:
You can't use decorators with WSGI.

mark-adams:
That's not an entirely unreasonable solution. I was trying to be flexible for the use case where somebody does have a path that has a component (like an ID) that is not predictable in advance.

As a further compromise, what if we made it take a list of strings or a callable which returns true if you skip; false if you don't. That way the library maintains the simple behavior you're looking for but gives advanced use cases an out if they require it.

atlassian_jwt_auth/frameworks/common/asap.py
```diff
@@ -14,6 +14,13 @@ def _process_asap_token(request, backend, settings):
     if token is None and not settings.ASAP_REQUIRED and (
             settings.ASAP_REQUIRED is not None):
         return
+
+    if request and settings.ASAP_EXCLUDE_PATHS:
```
dbaxa: If we want to do this I think this should be done in the various middleware classes or in a different function than this one.

PR #91, opened 2018-08-03
Sem-Ver: bugfix Reduce the time taken to generate a jwt by passing through a load private key instance to jwt.encode.

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/signer.py
+++ b/atlassian_jwt_auth/signer.py
@@ -3,6 +3,8 @@ import datetime
 import random
 
 import jwt
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
 
 from atlassian_jwt_auth import algorithms
 from atlassian_jwt_auth import key
@@ -24,6 +26,18 @@ class JWTAuthSigner(object):
         if self.lifetime > datetime.timedelta(hours=1):
             raise ValueError("lifetime, '%s',exceeds the allowed 1 hour max" %
                              (self.lifetime))
+
+    @property
+    def _private_key(self):
+        key_identifier, private_key_pem = self.private_key_retriever.load(
+            self.issuer)
+        if not isinstance(private_key_pem, bytes):
+            private_key_pem = private_key_pem.encode()
+        return serialization.load_pem_private_key(
+            private_key_pem,
+            password=None,
+            backend=default_backend()
+        )
 
     def _generate_claims(self, audience, **kwargs):
         """ returns a new dictionary of claims. """
@@ -50,7 +64,7 @@ class JWTAuthSigner(object):
             self.issuer)
         return jwt.encode(
             self._generate_claims(audience, **kwargs),
-            key=private_key_pem,
+            key=self._private_key,
             algorithm=self.algorithm,
             headers={'kid': key_identifier.key_id})
 
--- a/atlassian_jwt_auth/signer.py
+++ b/atlassian_jwt_auth/signer.py
@@ -18,6 +18,7 @@ class JWTAuthSigner(object):
         self.lifetime = kwargs.get('lifetime', datetime.timedelta(hours=1))
         self.algorithm = kwargs.get('algorithm', 'RS256')
         self.subject = kwargs.get('subject', None)
+        self._private_keys_cache = dict()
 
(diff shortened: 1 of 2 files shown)


dbaxa:
We may consider this a (slight) api break.

atlassian_jwt_auth/signer.py
```diff
@@ -50,7 +64,7 @@ def generate_jwt(self, audience, **kwargs):
             self.issuer)
         return jwt.encode(
             self._generate_claims(audience, **kwargs),
-            key=private_key_pem,
+            key=self._private_key,
```
dbaxa: This almost halves the time taken to encode a RS256 jwt (after the first access to `_private_key`).

mark-adams: Hmmm... I'm not getting the same results when I profile this function (master and your branch are essentially the same). I don't think the `@property` decorator does any sort of caching or memoization of the call; it is just syntatic sugar so you don't have to call the method with parenthesis.

I created a branch where we explicitly check for `self._private_key is None` and then populate it if it is not present. The benchmarks are about 50% lower with that change. I think that may have been what you were shooting for?

https://github.com/atlassian/asap-authentication-python/pull/92

dbaxa:
cc @jbunton-atlassian - I think that the change that is being made here would arguably break `FilePrivateKeyRetriever` because we would load a private key only once. Instead this code would need to be changed to retain a cache based upon `key_identifier`.

dbaxa:
I have updated the code to use a cache of private keys - to avoid the problem of having a private key retriever only be called once.

atlassian_jwt_auth/signer.py
```diff
@@ -25,6 +28,25 @@ def __init__(self, issuer, private_key_retriever, **kwargs):
             raise ValueError("lifetime, '%s',exceeds the allowed 1 hour max" %
                              (self.lifetime))
 
+    def _obtain_private_key(self, key_identifier, private_key_pem):
```
mark-adams: Nit: `obtain` is a little too fancy for me. Why not `get`?

dbaxa: I prefer `obtain` over `get` in this case because it isn't a simple access.

dbaxa:
Performance change -> 

Before this change
```
ec 3.3106096630217507
rsa 14.982443070970476
```

After this change
```
ec 2.271472151041962
rsa 8.497295178007334
```

Benchmarking script (based off @mark-adams 's found in https://github.com/atlassian/asap-authentication-python/pull/92) 
```
import timeit

import atlassian_jwt_auth
from cryptography.hazmat.backends import default_backend
from cryptography.hazmat.primitives.asymmetric import rsa, ec
from cryptography.hazmat.primitives.serialization import Encoding
from cryptography.hazmat.primitives.serialization import PrivateFormat
from cryptography.hazmat.primitives import serialization


def wrapper(func, *args, **kwargs):
    def wrapped():
        return func(*args, **kwargs)
    return wrapped


def sign(signer):
    signer.generate_jwt('audience')


def main():
    ec_private_key = ec.generate_private_key(
        ec.SECP256R1(), default_backend())

    rsa_private_key = rsa.generate_private_key(
        public_exponent=65537,
        key_size=2048,
        backend=default_backend()
    )

    ec_private_private_pem = ec_private_key.private_bytes(
        encoding=serialization.Encoding.PEM,
        format=serialization.PrivateFormat.TraditionalOpenSSL,
        encryption_algorithm=serialization.NoEncryption()
    )

    rsa_private_key_pem = rsa_private_key.private_bytes(
        Encoding.PEM, PrivateFormat.PKCS8, serialization.NoEncryption()
    )
    issuer = 'issuer'
    key_id = 'issuer/key'

    ec_signer = atlassian_jwt_auth.create_signer(
        issuer, key_id, ec_private_private_pem, algorithm='ES256'
    )
    rsa_signer = atlassian_jwt_auth.create_signer(
        issuer, key_id, rsa_private_key_pem, algorithm='RS256'
    )
    ec_w = wrapper(sign, ec_signer)
    rsa_w = wrapper(sign, rsa_signer)
    num = 10000
    print('ec', timeit.timeit(ec_w, number=num))
    print('rsa', timeit.timeit(rsa_w, number=num))


main()
```

PR #93, opened 2018-08-22
Reuse verifier in Backend

grantisu:
Sem-Ver: bugfix


atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -83,6 +83,10 @@ def get_asap_token(self, request):
 
     def get_verifier(self, settings=None):
         """Returns a verifier for ASAP JWT tokens"""
+        verifier = getattr(self, '_verifier', None)
```
dbaxa: This might be okay if `settings` is `None` but what about when `settings` is not `None`?

dbaxa:
A different approach has been taken.

PR #95, opened 2018-09-04
Fix super call issue in Django middleware

mbertrand80:


atlassian_jwt_auth/contrib/django/middleware.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, get_response=None):
                              'HTTP_X_ASAP_AUTHORIZATION')
 
     def process_request(self, request):
-        error_response = super(ProxiedAsapMiddleware, super).process_request(
```
mark-adams: Whoops!

PR #96, opened 2018-10-03
Allow specifying statuses for fetcher failover

grantisu:
The goal here is to allow using key servers with disjoint sets of keys, which is why 404 doesn't cause an immediate failure.

### DIFF
--- a/atlassian_jwt_auth/key.py
+++ b/atlassian_jwt_auth/key.py
@@ -123,9 +123,12 @@ class HTTPSMultiRepositoryPublicKeyRetriever(BasePublicKeyRetriever):
         repository locations based upon key ids.
     """
 
-    def __init__(self, key_repository_urls):
+    def __init__(self, key_repository_urls, failover_on=None):
         if not isinstance(key_repository_urls, list):
             raise TypeError('keystore_urls must be a list of urls.')
+        if failover_on is None:
+            failover_on = (404, 500, 502, 503, 504)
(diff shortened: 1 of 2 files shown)


dbaxa:
I remember having a discussion about this and my response at the time was "no let's not go this". The case where someone was seeing keys not being propagated was related to the dev environment iirc. Are you seeing one key repository 404 while another one isn't - if so let's try and get that fixed at an infrastructure level (or provide a means to opt into the behaviour you are desiring through this change - but not have it as the default)

grantisu:
The reason I need this is to allow a service running in production to talk to a service running in staging. Beyond that, handling split-brain in keyservers doesn't seem like a terrible idea; if a server *really* needs to make sure a key isn't fetched from elsewhere, it can use a 410.

atlassian_jwt_auth/key.py
```diff
@@ -123,9 +123,12 @@ class HTTPSMultiRepositoryPublicKeyRetriever(BasePublicKeyRetriever):
         repository locations based upon key ids.
     """
 
-    def __init__(self, key_repository_urls):
+    def __init__(self, key_repository_urls, failover_on=None):
         if not isinstance(key_repository_urls, list):
             raise TypeError('keystore_urls must be a list of urls.')
+        if failover_on is None:
```
dbaxa: I don't think we should have this "active" by default.

PR #98, opened 2019-02-27
HTTP Caching: Share cache across key retriever instances

erikvanzijst:
This change makes it so that when instantiating the HTTPSPublicKeyRetriever class
multiple times, they all share the same requests session with cache.

Previously, each instance would create its own session with isolated cache. Since
the library by default always creates a new instance of the ASAP_KEY_RETRIEVER_CLASS
class (e.g. common.backend.Backend#get_verifier), cached responses were never reused.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,10 @@
 CHANGES
 =======
 
+4.1.3
+
+* Changed HTTP caching session to be shared among HTTPSPublicKeyRetriever instances
+
 4.1.2
 -----
 
--- a/atlassian_jwt_auth/key.py
+++ b/atlassian_jwt_auth/key.py
@@ -72,6 +72,9 @@ class HTTPSPublicKeyRetriever(BasePublicKeyRetriever):
     """ This class retrieves public key from a https location based upon the
          given key id.
     """
+    # Use a static requests session, reused/shared by all instances of
+    # HTTPSPublicKeyRetriever:
+    _session = None
 
     def __init__(self, base_url):
         if base_url is None or not base_url.startswith('https://'):
@@ -80,12 +83,12 @@ class HTTPSPublicKeyRetriever(BasePublicKeyRetriever):
         if not base_url.endswith('/'):
             base_url += '/'
         self.base_url = base_url
-        self._session = self._get_session()
-
-    def _get_session(self):
-        session = requests.Session()
-        session.mount('https://', cachecontrol.CacheControlAdapter())
-        return session
+        if HTTPSPublicKeyRetriever._session is None:
+            HTTPSPublicKeyRetriever._session = requests.Session()
+            HTTPSPublicKeyRetriever._session.mount(
+                'http://', cachecontrol.CacheControlAdapter())
+            HTTPSPublicKeyRetriever._session.mount(
+                'https://', cachecontrol.CacheControlAdapter())
 
     def retrieve(self, key_identifier, **requests_kwargs):
         """ returns the public key for given key_identifier. """
--- a/atlassian_jwt_auth/tests/test_public_key_provider.py
+++ b/atlassian_jwt_auth/tests/test_public_key_provider.py
@@ -1,5 +1,7 @@
+import re
 import unittest
 
+import httptest
 import mock
 import requests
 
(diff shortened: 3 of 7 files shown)


erikvanzijst:
The mismatch between having per-instance session cache mounts and having the library re-instantiate the `HTTPSPublicKeyRetriever` every time means that projects using the embedded web framework decorators end up with a non-functioning cache, downloading the same keys over and over on every incoming ASAP request.

Bitbucket Cloud itself as well as Bitbucket hosted-storage contain custom hacks to work around this in most circumstances, but other projects (like PRLinks) not aware of these issues run without caching in production.

This should fix that for all users.

ChangeLog
```diff
@@ -1,6 +1,10 @@
 CHANGES
 =======
 
+4.1.3
```
erikvanzijst: How does the versioning work in this project? I couldn't find any version string outside this text file. I also don't see any automated PyPi deploys. Do you do PyPi uploads manually, @dbaxa ?

atlassian_jwt_auth/tests/test_public_key_provider.py
```diff
@@ -93,6 +97,50 @@ def test_retrieve_fails_with_forbidden_error(self, mock_get_method):
...
+        with httptest.testserver(wsgi) as server:
+
+            retriever = self.HTTPPublicKeyRetriever(server.url())
+            retriever.retrieve('example/eg')
+
+            retriever = self.HTTPPublicKeyRetriever(server.url())
+            retriever.retrieve('example/eg')
+
+            self.assertEqual(1, len(server.log()),
+                             msg='HTTP caching should suppress second GET')
```
erikvanzijst: This should prevent future regressions.

atlassian_jwt_auth/tests/test_public_key_provider.py
```diff
@@ -93,6 +97,50 @@ def test_retrieve_fails_with_forbidden_error(self, mock_get_method):
...
+    def test_http_caching(self):
+        """Asserts that our use of requests properly caches keys between
+        invocations across different `HTTPSPublicKeyRetriever` instances.
+        """
+        def wsgi(environ, start_response):
+            print(environ['PATH_INFO'])
+            start_response('200 OK', [
+                ('content-type', 'application/x-pem-file;charset=UTF-8'),
+                ('Cache-Control', 'public,max-age=300,stale-while-revalidate='
+                                  '300,stale-if-error=300'),
```
erikvanzijst: These are the actual cache control values as returned by the key servers.

ChangeLog
dbaxa: This project uses `pbr` which does the version bumping (somewhat) for us. I manually tweak the resulting changelog that pbr generates though.

setup.py
```diff
@@ -4,6 +4,8 @@
 
 setup(
     setup_requires=['pbr<=6.0.0'],
+    dependency_links=['https://bitbucket.org/atlassian/httptest/get/0.8.zip'
```
dbaxa: This needs / should be available from `pypi`.

ChangeLog
erikvanzijst: Cool, I'll leave that to you then.

setup.py
erikvanzijst: I wasn't aware it was published to PyPi (https://pypi.org/project/httptest/ is something entirely different), but I'll switch that!

atlassian_jwt_auth/key.py
```diff
@@ -72,6 +72,9 @@ class HTTPSPublicKeyRetriever(BasePublicKeyRetriever):
     """ This class retrieves public key from a https location based upon the
          given key id.
     """
+    # Use a static requests session, reused/shared by all instances of
+    # HTTPSPublicKeyRetriever:
+    _session = None
```
dbaxa: I'd prefer to not make such a change in this pull request. This is arguably a breaking behaviour / a change.

dbaxa: Can we create a `CacheControl` instance that we mount to sessions?

erikvanzijst: We could, though I'm not sure it's desirable to be creating new sessions for each retriever instance either. The session holds on the keepalive connections to reuse them the next time, and possibly other things.

I think the primary thinking here is to have the default behavior share the entire, single session between all instances of the HTTPSPublicKeyRetriever.

PR #99, opened 2019-02-27
Sem-Ver: api-break Re-use verifiers in the various middlewares and add an optional verifier argument to the _process_asap_token method.

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/frameworks/common/asap.py
+++ b/atlassian_jwt_auth/frameworks/common/asap.py
@@ -6,7 +6,7 @@ from atlassian_jwt_auth.exceptions import (
 )
 
 
-def _process_asap_token(request, backend, settings):
+def _process_asap_token(request, backend, settings, verifier=None):
     """ Verifies an ASAP token, validates the claims, and returns an error
     response"""
     token = backend.get_asap_token(request)
@@ -17,7 +17,8 @@ def _process_asap_token(request, backend, settings):
     try:
         if token is None:
             raise NoTokenProvidedError
-        verifier = backend.get_verifier(settings=settings)
+        if verifier is None:
+            verifier = backend.get_verifier(settings=settings)
         asap_claims = verifier.verify_jwt(
             token,
             settings.ASAP_VALID_AUDIENCE,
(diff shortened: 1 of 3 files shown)


mbertrand80:
Also, this caching doesn't help in practice. Many python based server applications recycle their workers every _n_ requests. This means we're still going to the keyserver way more often than is necessary or practical to get a public key. With this change, all of the internal applications that rely on this package will still have to apply the same custom hacks to get stable caching, retry, and fallback.

mbertrand80:
I have a private implementation of the `MultiRepoPublicKeyRetriever` and `HTTPSPublicKeyRetriever` that provides session caching, utilizes Django's caching (if available), and fallback to any number of keystores provided. I can try to get that put into a Pull request soon.

dbaxa:
@atl-gmathews I am not overly happy with the design of `Backend`, I prefer to treat my class instances as being more immutable - so I would change the various `Backend` to support being given settings but would not allow providing specific settings in the `get_verifier` method. This would then allow for `get_verifier` to return/use a cached verifier.

erikvanzijst:
> This caching doesn't help in practice. Many python based server applications recycle their workers every _n_ requests. This means we're still going to the keyserver way more often than is necessary or practical to get a public key.

I don't think it's as bad as is sounds TBH. Yes, a cold worker process will incur a key download the first time, but:

* Not all Python services use the one process per request model (hostedstorage doesn't, PRLinks doesn't)
* The current key server's cache control expiry is only 5 minutes and so even long-lived processes have to re-download the keys more than I would have assumed

erikvanzijst:
FWIW I took yet another approach that is lower down than the decorators or middlewares and changes the actual HTTPSPubliKeyRetriever's use of requests sessions: https://github.com/atlassian/asap-authentication-python/pull/98

erikvanzijst:
> * The current key server's cache control expiry is only 5 minutes and so even long-lived processes have to re-download the keys more than I would have assumed

Note that the key server seems to have been aware of this issue and turned to the use of a non-standardized cache-control directive (`stale-while-revalidate`) to try and trigger asynchronous, background re-fetched.

However, the fact that this directive was never actually adopted in any published RFC means that it doesn't actually work for us: https://github.com/ionrock/cachecontrol/issues/204

PR #100, opened 2019-03-01
Sem-Ver: api-break Share request sessions across key retriever instances so as to use a common cache.

dbaxa:
This change makes it so that when instantiating the HTTPSPublicKeyRetriever class
    multiple times, they all share the same requests session with cache.

    Previously, each instance would create its own session with isolated cache. Since
    the library by default always creates a new instance of the ASAP_KEY_RETRIEVER_CLASS
    class (e.g. common.backend.Backend#get_verifier), cached responses were never reused.


dbaxa:
@erikvanzijst  I committed this as you as the author so as to give you credit.

PR #105, opened 2019-09-29
Added an example on how to generate token using data uri private key

itsrifat:
Might be useful for people looking to use base64 encoded data uri


README.rst
```diff
@@ -48,6 +48,20 @@ Each time you call ``generate_jwt`` this will find the latest active key file (e
...
 
+To create a JWT using a data uri
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: python
+
+    import atlassian_jwt_auth
+    from atlassian_jwt_auth.key import DataUriPrivateKeyRetriever
+
+    pem_id, private_key_pem = DataUriPrivateKeyRetriever('Your base64 encoded data uri').load('issuer/key')
```
dbaxa: @itsrifat - the `.load` method takes an `issuer` and not a `key identifier` - so arguably this should be `DataUriPrivateKeyRetriever('Your base64 encoded data uri').load('issuer')`.

PR #109, opened 2019-10-23
Sem-Ver: feature Add support for python 3.7

dbaxa:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
+  - "3.7"
 install:
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 || $TRAVIS_PYTHON_VERSION == 3.6 ]]; then pip install aiohttp==2.0.6 asynctest==0.10.0; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,8 +4,9 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
+  - "3.7"
 install:
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 || $TRAVIS_PYTHON_VERSION == 3.6 ]]; then pip install aiohttp==2.0.6 asynctest==0.10.0; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 || $TRAVIS_PYTHON_VERSION == 3.6 || $TRAVIS_PYTHON_VERSION == 3.7 ]]; then pip install aiohttp==3.6.2 asynctest==0.13.0; fi
```
dbaxa: We need a newer version of `aiohttp` to support python 3.7 & while I was here I also updated the version of `asynctest` test.

atlassian_jwt_auth/contrib/tests/aiohttp/test_public_key_provider.py
```diff
@@ -21,7 +21,7 @@ def _get_session(self):
 
         resp = session.get.return_value
         resp.headers = CIMultiDict({"content-type": PEM_FILE_TYPE})
-        resp.text.return_value = 'i-am-a-public-key'
+        resp.text = CoroutineMock(return_value='i-am-a-public-key')
```
dbaxa: `resp.text` is a coroutine (https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.text).

PR #111, opened 2020-02-23
Support disabling the jti uniqueness check in the various frameworks

dbaxa:


atlassian_jwt_auth/frameworks/wsgi/tests/test_wsgi.py
```diff
@@ -63,19 +63,42 @@ def test_request_with_valid_token_is_allowed(self):
             'client-app/key01', self._private_key_pem
         )
         body, resp_info, environ = self.send_request(token=token)
-        assert resp_info['status'] == '200 OK'
-        assert 'ATL_ASAP_CLAIMS' in environ
+        self.assertEqual(resp_info['status'], '200 OK')
+        self.assertIn('ATL_ASAP_CLAIMS', environ)
+
+    @unittest.expectedFailure
```
dbaxa: @mark-adams I haven't been able to determine why the wsgi middleware doesn't behave like the other middleware with regards to respecting the `ASAP_CHECK_JTI_UNIQUENESS` setting.

dbaxa: I ended up working out that I needed to persist the wsgi application when sending more than one request to fix the test.

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -38,7 +42,10 @@ class Backend():
         'ASAP_VALID_ISSUERS': None,
 
         # Enforce that the ASAP subject must match the issuer
-        'ASAP_SUBJECT_SHOULD_MATCH_ISSUER': None
+        'ASAP_SUBJECT_SHOULD_MATCH_ISSUER': None,
+
+        # Enforce that tokens have a unique JTI
+        'ASAP_CHECK_JTI_UNIQUENESS': None,
```
dbaxa: I am thinking that for historical reasons it makes more sense to default to the JTI not being checked by default.

atlassian_jwt_auth/frameworks/common/backend.py
```diff
@@ -85,14 +92,19 @@ def get_verifier(self, settings=None):
         """Returns a verifier for ASAP JWT tokens"""
         if settings is None:
             settings = self.settings
+        return self._get_verifier(settings)
 
+    @lru_cache(maxsize=130)
```
dbaxa: We can lower this value.

atlassian_jwt_auth/frameworks/common/backend.py
jbirch-atlassian: I think that's fine; the spec calls out that if people are enforcing it, they need to document it — thus this is the safe default:

> The resource server MAY reject a token if the token nonce (“jti”) has been previously seen by the resource server in another request. If the resource server decides to implement duplicate detection, it MUST explicitly document that behaviour.

PR #119, opened 2020-02-27
Sem-Ver: feature Add logging to the framework asap token checking code.

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/frameworks/common/asap.py
+++ b/atlassian_jwt_auth/frameworks/common/asap.py
@@ -1,3 +1,5 @@
+import logging
+
 from jwt.exceptions import InvalidIssuerError, InvalidTokenError
 
 from atlassian_jwt_auth.exceptions import (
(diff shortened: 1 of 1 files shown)


atlassian_jwt_auth/frameworks/common/asap.py
```diff
@@ -11,6 +13,7 @@
 def _process_asap_token(request, backend, settings, verifier=None):
     """ Verifies an ASAP token, validates the claims, and returns an error
     response"""
+    logger = logging.getLogger('asap')
```
hubertgrzeskowiak: Do we call `logging.basicConfig()` anywhere? There might be inconsistencies if it's not called early. See https://stackoverflow.com/questions/7016056/python-logging-not-outputting-anything/56144390#56144390

hubertgrzeskowiak: Is `asap` enough of context for people using this library to connect it to `atlassian-jwt-auth`? (btw, it's kinda confusing that the pip package and the repo name don't match at all)

PR #124, opened 2020-05-21
Wip allow specifying status for fetcher failover

dbaxa:


dbaxa:
cc @grantisu grantisu

PR #130, opened 2020-07-09
Sem-Ver: api-break Optimise the standard requests session based public key retrieval by obtaining proxy information from the environment once and setting session.trust_env to False.  As a result of this change retrieving the same public key, with caching enabled, 10,000 times takes ~ 6 seconds instead of ~ 14 seconds.

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/key.py
+++ b/atlassian_jwt_auth/key.py
@@ -9,6 +9,7 @@ import cachecontrol
 import cryptography.hazmat.backends
 import jwt
 import requests
+import requests.utils
 from cryptography.hazmat.primitives import serialization
 from requests.exceptions import RequestException, ConnectionError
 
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/key.py
```diff
@@ -88,6 +89,9 @@ def __init__(self, base_url):
     def _get_session(self):
         if HTTPSPublicKeyRetriever._class_session is None:
             session = cachecontrol.CacheControl(requests.Session())
+            session.proxies = requests.utils.get_environ_proxies(
+                self.base_url)
+            session.trust_env = False
```
mark-adams, line 94: Is there a difference in behavior between `False` and the default of `None` here or are you just being explicit?

dbaxa: I am not sure, I thought it would be safer to go with `False` though.

PR #132, opened 2020-07-09
Sem-Ver: feature Implement public key object caching.

dbaxa:


dbaxa:
This follows on from https://github.com/atlassian/asap-authentication-python/pull/130. The public key object caching reduces the difference between verifying a jwt `10,000` times - where a public key retriever provides an already loaded public key object instance & where the public key retriever provides a non-loaded instance by ~ 0.500 seconds.

PR #133, opened 2020-07-09
Sem-Ver: feature Implement public key object caching

dbaxa:


dbaxa:
@jbirch-atlassian  - do you have any views on this - does it look good to you ? :-)

atlassian_jwt_auth/verifier.py
```diff
@@ -31,15 +40,30 @@ def verify_jwt(self, a_jwt, audience, leeway=0, **requests_kwargs):
...
             key_identifier, **requests_kwargs)
 
+    @lru_cache(maxsize=10)
+    def _load_public_key(self, public_key, algorithm):
+        """ Returns a public key object instance given the public key and
+            algorithm.
+        """
+        if algorithm not in self.algorithms:
+            raise InvalidAlgorithmError(
+                'The specified alg value is not allowed')
```
hubertgrzeskowiak, line 62: Can we make the error tell the reader which algos are allowed?

atlassian_jwt_auth/verifier.py
```diff
@@ -31,15 +40,30 @@ def verify_jwt(self, a_jwt, audience, leeway=0, **requests_kwargs):
         """
         key_identifier = key._get_key_id_from_jwt_header(a_jwt)
         public_key = self._retrieve_pub_key(key_identifier, requests_kwargs)
-
+        public_key_obj = public_key
```
hubertgrzeskowiak, line 43: nit: this alias is rather confusing

jbirch-atlassian:
Broadly, it looks fine to me. It's not immediately clear why the objectification needs be lru-cached? I'd naively have thought the biggest impact would have come from lru-caching the fetch of the public key material. I think I'd need to look at more than just this diff to understand more fully what's going on here.

jbirch-atlassian (approved):

atlassian_jwt_auth/verifier.py
dbaxa: Yes, although they are part of the s2s asap spec and iirc this error matches what the pyjwt library produces.

atlassian_jwt_auth/verifier.py
dbaxa: Agreed.

atlassian_jwt_auth/verifier.py
```diff
@@ -31,15 +40,30 @@ def verify_jwt(self, a_jwt, audience, leeway=0, **requests_kwargs):
...
+    def _load_public_key(self, public_key, algorithm):
+        """ Returns a public key object instance given the public key and
+            algorithm.
+        """
+        if algorithm not in self.algorithms:
+            raise InvalidAlgorithmError(
+                'The specified alg value is not allowed')
+        py_jws = jwt.api_jws.PyJWS(algorithms=self.algorithms)
+        alg_obj = py_jws._algorithms[algorithm]
+        return alg_obj.prepare_key(public_key)
```
dbaxa, line 65: Confusingly we need to go via `jws` and can't make use of `jwt` directly here.

dbaxa:
> Broadly, it looks fine to me. It's not immediately clear why the objectification needs be lru-cached? I'd naively have thought the biggest impact would have come from lru-caching the fetch of the public key material. I think I'd need to look at more than just this diff to understand more fully what's going on here.

It doesn't have to be, but there is a performance gain in caching the object & if the underlying key material hasn't changed there & we have an existing object available we may as well use it.

tildedave:
I may be misunderstanding something but it seems that right now we're not using any kind of caching for keys in the asyncio flow.  Should this PR create an interface that the asyncio HTTPSPublicKeyRetriever can hook into?  If it's straightforward happy to look at creating a patch for this.

atlassian_jwt_auth/verifier.py
tildedave, line 43: Seems like this logic could be in a method that the asyncio verifier could call which would give the benefits of caching for this flow as well.

tildedave (approved):

atlassian_jwt_auth/verifier.py
tildedave: Hrm ... the actual thing that needs to be cached here is the `await public_key` call in the asyncio/verifier.py.  Maybe the verifier should just be caching based on the key_identifier.  But I'm not certain if this is done on purpose or not.

dbaxa: > Seems like this logic could be in a method that the asyncio verifier could call which would give the benefits of caching for this flow as well.

Good point, I will adjust this pr.

PR #138, opened 2021-01-14
Sem-Ver: feature Add support for using PyJWT 2.0.x

dbaxa:
These changes provide support for using PyJWT 1.x and 2.0.x.

### DIFF
--- a/atlassian_jwt_auth/signer.py
+++ b/atlassian_jwt_auth/signer.py
@@ -72,11 +72,14 @@ class JWTAuthSigner(object):
             self.issuer)
         private_key = self._obtain_private_key(
             key_identifier, private_key_pem)
-        return jwt.encode(
+        token = jwt.encode(
             self._generate_claims(audience, **kwargs),
             key=private_key,
             algorithm=self.algorithm,
             headers={'kid': key_identifier.key_id})
+        if isinstance(token, str):
+            token = token.encode('utf-8')
+        return token
 
 
 class TokenReusingJWTAuthSigner(JWTAuthSigner):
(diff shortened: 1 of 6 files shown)


atlassian_jwt_auth/signer.py
```diff
@@ -72,11 +72,14 @@ def generate_jwt(self, audience, **kwargs):
...
             key_identifier, private_key_pem)
-        return jwt.encode(
+        token = jwt.encode(
             self._generate_claims(audience, **kwargs),
             key=private_key,
             algorithm=self.algorithm,
             headers={'kid': key_identifier.key_id})
+        if isinstance(token, str):
+            token = token.encode('utf-8')
+        return token
```
dbaxa, line 82: To remain consistent.

dbaxa:
Related: https://github.com/jpadilla/pyjwt/blob/master/CHANGELOG.rst

atlassian_jwt_auth/signer.py
jbirch-atlassian: Ah really, PyJWT changed between bytes and str between versions?

OOC, if and when support only PyJWT 2+, will we also change this to be in line with the types it returns?

jbirch-atlassian (approved):
Seems cool to me. Sorta annoying we have to repeat ourselves to be compatible with both versions — is there any particular reason we didn't just bump to PyJWT 2+?

atlassian_jwt_auth/signer.py
dbaxa: Possibly, although it might result in un-needed churn :shrug: .

dbaxa:
> Seems cool to me. Sorta annoying we have to repeat ourselves to be compatible with both versions — is there any particular reason we didn't just bump to PyJWT 2+?

For compatibility reasons and to not need a major version bump so soon after performing a major version bump :-) .

PR #139, opened 2021-02-25
Sem-Ver: api-break upgrade cryptography from using a version >= 3.3.1 < 3.4.0 to use a version >= 3.3.2 and < 3.5.0

dbaxa:


dbaxa:
We will hold off on merging this for the moment due to the rust changes & some might be using alpine, see https://github.com/pyca/cryptography/blob/7e30277483f5556ba05ab57edec11f0b8eebbb6e/docs/installation.rst for details.

PR #144, opened 2021-06-25
Sem-Ver: bugfix The aiohttp public key retriever incorrectly provided a proxies argument when it needed to provide a proxy argument.

dbaxa:


dbaxa:
This was reported by  @tildedave

tildedave (approved):
Thanks for the fix!

PR #157, opened 2022-03-31
Sem-Ver: bugfix Move away from using asynctest in python versions where unittest async utils are available (3.8 and above).

dbaxa:

### DIFF
--- a/atlassian_jwt_auth/contrib/tests/aiohttp/test_public_key_provider.py
+++ b/atlassian_jwt_auth/contrib/tests/aiohttp/test_public_key_provider.py
@@ -2,8 +2,14 @@ import os
 from unittest import mock
 
 import aiohttp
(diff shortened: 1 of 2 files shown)


atlassian_jwt_auth/contrib/aiohttp/key.py
```diff
@@ -43,10 +43,12 @@ async def _retrieve(self, url, requests_kwargs):
         requests_kwargs = self._convert_proxies_to_proxy_arg(
             url, requests_kwargs)
         try:
-            resp = await self._session.get(url, headers={'accept':
-                                                         PEM_FILE_TYPE},
-                                           **requests_kwargs)
-            resp.raise_for_status()
+            resp = await self._session.get(
+                url,
+                raise_for_status=True,
```
dbaxa, line 48: Let aiohttp call this for us (simplifies testing).

